PR #1, opened 2015-08-31
Initial plugin

EliAndrewC:
Must be merged after https://github.com/magfest/ubersystem/pull/1434

Right now this contains two things:

1) The original `Event` code for making a schedule and such.  This is all unchanged.

2) A new panel application form.  There's no way for admins to approve/decline these applications yet; that will need to be implemented later.

I'll leave comments inline, as usual.


panels/automated_emails.py
```diff
@@ -0,0 +1,6 @@
+from panels import *
+
+AutomatedEmail.extra_models[PanelApplication] = lambda session: session.query(PanelApplication).all()
+
+AutomatedEmail(PanelApplication, 'Your {EVENT_NAME} Panel Application Has Been Received', 'panel_app_confirmation.txt',
+               lambda app: True)
```
EliAndrewC, line 6: This is the first time we've wanted to define an `AutomatedEmail` with literally no `filter` parameter.  Perhaps we should change `filter` to have a default value of `lambda m: True`.

panels/config.py
```diff
@@ -0,0 +1,12 @@
+from panels import *
+
+panels_config = parse_config(__file__)
+c.include_plugin_config(panels_config)
```
EliAndrewC, line 4: As mentioned in the other PR at https://github.com/magfest/ubersystem/pull/1434 the `c.include_plugin_config()` method was extended to also look for `[enums]` and `[dates]` sections.

panels/configspec.ini
```diff
@@ -0,0 +1,32 @@
+
```
EliAndrewC, line 1: I think the comments explain everything, but let me know if there are any questions.

panels/model_checks.py
```diff
@@ -0,0 +1,44 @@
+from panels import *
```
EliAndrewC, line 1: The `Event` stuff is moved here from the core repo.  The rest is new, and hopefully seems straightforward.

panels/site_sections/panel_app_management.py
```diff
@@ -0,0 +1,7 @@
+from panels import *
+
+
+@all_renderable(c.STUFF)
+class Root:
+    def index(self, session):
+        return {'apps': session.query(PanelApplication).order_by('applied').all()}
```
EliAndrewC, line 7: As I mentioned, we'll need to eventually add some admin code to allow us to actually manage the applications.

panels/site_sections/schedule.py
```diff
@@ -0,0 +1,194 @@
+from panels import *
```
EliAndrewC, line 1: This file was moved here from the core repo and is otherwise unchanged (save for the single `import` line above).

panels/tests/test_event.py
```diff
@@ -0,0 +1,31 @@
+import pytest
```
EliAndrewC, line 1: These tests were moved here from the core repo.  I didn't write any new tests.  Someday in the far future we won't be so busy implementing new functionality that we'll actually have time to write tests.  Someday.

panels/models.py
```diff
@@ -0,0 +1,81 @@
...
+    def minutes(self):
+        return (self.duration or 0) * 30
+
+    @property
+    def start_slot(self):
+        if self.start_time:
+            return int((self.start_time_local - c.EPOCH).total_seconds() / (60 * 30))
+
+
+class AssignedPanelist(MagModel):
```
kitsuta, line 35: It's a little unusual to see a relational table in a postgres DB - what was the reason for doing this instead of a many-to-many relationship?

panels/templates/emails/panel_app_confirmation.txt
```diff
@@ -0,0 +1,5 @@
+{{ panelapplication.submitter.first_name }},
+
+We've recieved your application for a {{ c.EVENT_NAME }} panel.
```
kitsuta, line 3: It'd be nice if this was `c.EVENT_NAME_AND_YEAR`.

panels/templates/panel_applications/index.html
```diff
@@ -0,0 +1,246 @@
+{% extends "preregistration/preregbase.html" %}
+{% block title %}Panel Applications{% endblock %}
+{% block backlink %}{% endblock %}
+{% block content %}
+
+<script>
+    var showOrHideHeldBefore = function () {
```
kitsuta, line 7: We have a lot of these kinds of functions, both here and in core - it might be worthwhile to make a utility function that takes the field-to-hide name and the condition instead so as to reduce clutter.

panels/models.py
EliAndrewC, line 35: Excellent question!  As of right now, there's really no compelling reason to do it this way.  My expectation when I original defined this model years ago was that I'd be adding new columns to it.  Kind of like how `Shift` is basically a many-to-many relationship between `Attendee` and `Job` but it has enough other attributes (like whether or not it was worked and what rating was given) that we made it an independent model like this one.

I'm still thinking that we'll end up adding new columns to this model as we make our scheduling code more sophisticated... though offhand I can't think of what that might be, so I might just be needlessly planning ahead for something that will never happen.

panels/templates/emails/panel_app_confirmation.txt
EliAndrewC, line 3: Good idea.  As you can probably tell, this is just a placeholder; the panels folks say that they'll send me the text of the real email later this week.  When that happens I'll make sure to use `c.EVENT_NAME_AND_YEAR`.

panels/templates/panel_applications/index.html
EliAndrewC, line 7: Hmm, that's probably a good idea, since I do hate having lots of functions like this which are so similar.  From thinking about what the function signature would look like though, this is probably the best I can come up with for how you'd call the utility method if it was fully generic:

``` javascript
makeShowOrHider({
    event: 'click',
    source: 'held_before',
    target: 'past_attendance',
    condition: function (field) {
        return field.prop('checked');
    }
});
```

There are probably some ways we could make that shorter.  For example, if the function assumes that you're designating a checkbox, then we could instead just say

``` javascript
makeShowOrHider({source: 'held_before', target: 'past_attendance'});
```

and then have some extra parameters we can use if it's a dropdown:

``` javascript
makeShowOrHider({
    type: 'dropdown',
    source: 'presentation',
    target: 'other_presentation':
    show_if: {{ c.OTHER }}
});
```

How does that all look?  (In case it's not clear, you'd just call this at the `<script></script>` level and then you wouldn't need to define the function, or call the function and set the handler in the `$(function () {  })` block.)

panels/models.py
kitsuta, line 35: Ah okay, gotcha.

PR #2, opened 2015-09-01
Add database migrations for panels plugin

kitsuta:


EliAndrewC:
It looks like #3 already covers the comments I would have made here.

PR #3, opened 2015-09-01
Fix panels migration to cover tables in core

kitsuta:
I mistakenly created the migration without removing the tables that were already in core first; I've manually added those changes in.


binary1230:
is this outdated now? can we close it?

kitsuta:
Whoops, yes.

PR #4, opened 2015-09-06
added json dump to public schedule

EliAndrewC:
As requested for external displays, since json is way easier to work with than XML.


EliAndrewC:
Merging this now so people can start testing with it tonight.

PR #6, opened 2015-10-22
Panel voting

EliAndrewC:
I had a couple of very brief discussions with the panel folks about this, and they'd like a voting system, albeit a simpler one than what MIVS requires.

I'm going to hold off on merging this until I have a chance to talk it over with the panels folks.

### DIFF
--- a/panels/configspec.ini
+++ b/panels/configspec.ini
@@ -40,3 +40,7 @@ movie = string(default="Media presentation (like a movie)")
 performance = string(default="Scripted or Improv Live Performance")
 participation = string(default="Audience participation (like a game)")
 other = string(default="Other")
+
+[[panel_vote]]
+accept = string(default="Accept")
+decline = string(default="Reject")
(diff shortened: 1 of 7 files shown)


panels/configspec.ini
```diff
@@ -40,3 +40,7 @@ movie = string(default="Media presentation (like a movie)")
 performance = string(default="Scripted or Improv Live Performance")
 participation = string(default="Audience participation (like a game)")
 other = string(default="Other")
+
+[[panel_vote]]
+accept = string(default="Accept")
+decline = string(default="Reject")
```
EliAndrewC, line 50: I made this an enum instead of a boolean since I figured they might want more than just yes/no in the future.

panels/models.py
```diff
@@ -1,4 +1,4 @@
-from panels import *    
+from panels import *
```
EliAndrewC, line 1: Whitespace fix.

PR #7, opened 2016-01-03
added ability to approve and decline applications

EliAndrewC:
I had to run the following command as part of this on both the staging and production servers:

``` sql
ALTER TABLE panel_application ADD COLUMN status INTEGER NOT NULL DEFAULT 196944751;
```


EliAndrewC:
Merging now so I can email the panel folks about it tonight.

PR #14, opened 2016-05-11
Gracefully handle missing panel rooms in INI

binary1230:
The panels plugin keeps track of what it calls "panel rooms"

These are supposed to also exist in "event_locations" in the main INI, but when they don't, we were crashing instead of saying something useful.  It took me a while to figure out why this was happening both with Magstock and MagLabs, so, doing two things here:

 1) Log a warning when a problem is detected that tells you where to go to fix it
 2) Gracefully handle the situation without crashing, so this doesn't stop the app from starting up

### DIFF
--- a/panels/config.py
+++ b/panels/config.py
@@ -15,7 +15,7 @@ invalid_panel_rooms = [room for room in c.PANEL_ROOMS if not getattr(c, room.upp
 
 for room in invalid_panel_rooms:
     log.warning('panels plugin: panels_room config problem: '
-                'Ignoring "' + room.upper() + '" because it was not also found in [[event_location]] section.')
+                'Ignoring {!r} because it was not also found in [[event_location]] section.'.format(room.upper()))
 
 c.PANEL_ROOMS = [getattr(c, room.upper()) for room in c.PANEL_ROOMS if room not in invalid_panel_rooms]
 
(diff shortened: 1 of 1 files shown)


panels/config.py
```diff
@@ -11,7 +11,13 @@
...
 c.EVENT_OPEN   = {'colspan': 1}
 
-c.PANEL_ROOMS = [getattr(c, room.upper()) for room in c.PANEL_ROOMS]
+invalid_panel_rooms = [room for room in c.PANEL_ROOMS if not getattr(c, room.upper(), None)]
+
+for room in invalid_panel_rooms:
+    log.warning('panels plugin: panels_room config problem: '
+                'Ignoring "' + room.upper() + '" because it was not also found in [[event_location]] section.')
+
+c.PANEL_ROOMS = [getattr(c, room.upper()) for room in c.PANEL_ROOMS if room not in invalid_panel_rooms]
```
binary1230, line 20: @EliAndrewC let me know if you think there might be a more compact way to do this that's also as readable, I couldn't figure out how to get a list comprehension to exclude an invalid entry while also emitting a warning message on items we wanted to exclude.

binary1230:
this PR btw is not really necessary, it will just make things easier the next time we have to mess around with bringing this plugin online and forget what half the INI settings are :)

panels/config.py
EliAndrewC, line 20: I think this is fine as-is.  Though for log messages you can say

```
log.warning('I want {!r} to be quoted', room.upper())
```

and then you don't have to muck around with `"' + room.upper() + '"`

binary1230:
integrated Eli's formatting suggestion, merging

PR #15, opened 2016-06-27
convert templates from django to jinja2 - panels

binary1230:
WIP, but pretty much ready. merge after https://github.com/magfest/ubersystem/pull/1888

Convert all templates to use JInja2 instead of Django.


binary1230:
+1

I would bet though that the schedule page needs some careful testing due to its extensive date/time formatting and the differences between how django and jinja handle that.  However, for now, I would suggest we ignore it and fix it later for the sake of making sure our branches don't diverge more

PR #17, opened 2016-08-07
updated panel application deadline for this year

EliAndrewC:
In the long run this should be puppetized.  In the short run this is how we update the date for this year.


EliAndrewC:
Gonna merge this in now since the panels folks would like to look at this.

PR #18, opened 2016-08-07
updated panel application response date for this year

EliAndrewC:
Additional fix.  Will eventually be puppetized, but for now this is how things are.


EliAndrewC:
Gonna merge this in now since it's a simple text change.

PR #19, opened 2016-08-07
removed panel voting and added new fields to application form

EliAndrewC:
Three changes, requested by the panel folks:

1) Eliminate the voting component.  This was never used, and is kind of clunky anyway.  There's some discussion of better ways to go about this, but for not it's just not useful.

2) Moved a couple of fields around on the form.

3) Updated the default deadline (eventually we ned to puppetize that).

4) Added a new field for the applicant to indicate what tech they're bringing, e.g. a Macbook or a Windows laptop or a game console, etc.  This required me to run the following, which I already did on the labs and supermag servers:

``` sql
ALTER TABLE panel_application ADD COLUMN panelist_bringing TEXT NOT NULL DEFAULT '';
```

Though since those tables were empty I could have also just dropped them and let the system re-create them, but this was easier since the PR hasn't been accepted yet.

### DIFF
--- a/development-defaults.ini
+++ b/development-defaults.ini
@@ -4,5 +4,5 @@ expected_response = "December 2016"
 event_location = "the Gaylord National Resort and Convention Center in National Harbor, Maryland"
 
 [dates]
-panel_app_deadline = "2016-10-30"
+panel_app_deadline = "2016-10-31"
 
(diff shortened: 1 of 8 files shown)


binary1230:
looked through it without testing it, looks sane to me.  if this is ready to rock and satisfies what the panels dept folks are asking for, we can deploy.  +1

PR #31, opened 2016-09-02
allow csv of schedule to be exported

daredoes:


binary1230:
need to keep the sort of c.ORDERED_EVENT_LOCS

EliAndrewC:
+1 (belatedly)

PR #32, opened 2016-09-09
Add redirect if alt_schedule_url is set

kitsuta:
For MAGLabs this year, we wanted to simply disable the schedule permanently and use Guidebook. However, this can be confusing for people who are used to accessing the schedule URL. This now gives events the ability to redirect to another URL instead.

### DIFF
--- a/development-defaults.ini
+++ b/development-defaults.ini
@@ -3,6 +3,8 @@ expected_response = "December 2016"
 
 event_location = "the Gaylord National Resort and Convention Center in National Harbor, Maryland"
 
+alt_schedule_url = "https://guidebook.com/guide/77688/schedule/"
+
 [dates]
 panel_app_deadline = "2016-10-31"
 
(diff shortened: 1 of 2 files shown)


binary1230:
i think it would make a bit more sense for it to be:

```
# check if allowed to view schedule before it's public
if c.HIDE_SCHEDULE and not AdminAccount.access_set() and not cherrypy.session.get('staffer_id'):
   return "The " + c.EVENT_NAME + " schedule is being developed and will be made public when it's closer to being finalized."

# they're allowed to view it so show it to them
if c.ALT_SCHEDULE_URL:
    raise HTTPRedirect(c.ALT_SCHEDULE_URL)

# ---- rest of code ----
```

kitsuta:
My thinking was that events would want to set "hide schedule" to true when they don't want to show RAMS' version of the schedule, and then in that case, they could _either_ show the 'not ready' message (if they just have no schedule) or redirect to an alternate schedule like Guidebook.

kitsuta:
We talked in Slack about it and decided that it was better if events were able to set up a schedule URL without having it immediately 'go live,' so now hide_schedule always hides the schedule even if the alt_schedule_url is set.

binary1230:
+1 looks great

Issue #49, opened 2016-09-14
Add panel name to application confirmation text


nickthenewbie1:
In this file: https://github.com/magfest/panels/blob/master/panels/templates/emails/panel_app_confirmation.txt

Please modify this sentence's text accordingly, with the appropriate data filled in:

"Your application for the panel '<PANEL NAME HERE>' has been received and is now being reviewed."

binary1230:
fixed by #50

PR #50, opened 2016-09-14
fix-#49

daredoes:
Adds panel name to email text as stated in issue #49


binary1230:
I think this looks right based on other areas of the code, we should review the emails going out to make sure it looks ok before approving them

+1 on this.

binary1230 (approved):
looks correct to me, after merge, please review any emails generated before approving them.

PR #59, opened 2017-01-01
added CSV export for panelist tech needs

EliAndrewC:
Clint and I talked a few days ago about this.  They'd like to be able to have a spreadsheet of all the tech needs for each of the panels broken down by location and time whichthey can give to tech ops to make sure that they know what's needed.

### DIFF
--- a/panels/site_sections/schedule.py
+++ b/panels/site_sections/schedule.py
@@ -244,6 +244,7 @@ class Root:
                           if a.paid == c.HAS_PAID or a.paid == c.PAID_BY_GROUP and a.group and a.group.amount_paid]
         }
 
+    @unrestricted
     def panelist_schedule(self, session, id):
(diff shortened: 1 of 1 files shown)


panels/site_sections/schedule.py
```diff
@@ -244,6 +244,7 @@ def panelists_owed_refunds(self, session):
                           if a.paid == c.HAS_PAID or a.paid == c.PAID_BY_GROUP and a.group and a.group.amount_paid]
         }
 
+    @unrestricted
```
EliAndrewC, line 247: This is an unrelated fix; I realized that I forgot to make the page that we will eventually email out to our panelists showing their schedule unrestricted, which would have prevented them from being able to see it!

kitsuta (approved):
A little OMHBITY

PR #64, opened 2017-01-02
Restore ability to view internal uber schedule

daredoes:
To be merged with PR in Ubersystem.

This will separate the guidebook and internal schedule into two functions, allowing staffers to view the previously used internal one. Some prefer it.

### DIFF
--- a/panels/site_sections/schedule.py
+++ b/panels/site_sections/schedule.py
@@ -9,9 +9,6 @@ class Root:
         if c.HIDE_SCHEDULE and not AdminAccount.access_set() and not cherrypy.session.get('staffer_id'):
             return "The " + c.EVENT_NAME + " schedule is being developed and will be made public when it's closer to being finalized."
 
-        if c.ALT_SCHEDULE_URL:
-            raise HTTPRedirect(c.ALT_SCHEDULE_URL)
-
         schedule = defaultdict(lambda: defaultdict(list))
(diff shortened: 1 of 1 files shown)


kitsuta:
This is cool and I totally get that some admins don't wanna be redirected to guidebook, but can we actually make this only happen for admins? Forcing attendees to the guidebook version is (AFAIK) ideal behavior.

EliAndrewC:
Agreed.  That page has *horrible* performance.  The non-guidebook schedule display page is probably the oldest bunch of unrefactored code in the entire codebase.  If you read through it, you can see it doing a bunch of computationally-intensive stuff that doesn't even apply to how we do things now, e.g. checking for how many simultaneous events are in each room at a time (we no longer allow this).

We should therefore only allow it for logged-in admins at most.  If we want to allow it at all, which I'm kind of down on, since admins can just view https://super2017.uber.magfest.org/uber/schedule/edit

binary1230:
Yea btw this was a request from CY because he wanted the option to view the old schedule because his computer doesn't like the guidebook format, and he liked the old schedule format.  Agreed that this should be a logged-in-user thing.

kitsuta (approved):

kitsuta:
Awesome! +11111

EliAndrewC:
+1

PR #71, opened 2017-01-06
Fixes #60

### LINKED ISSUE #60: As an admin, I'd like to delete applicants from a panel
We sometimes end up with applicants who shouldn't have badges created for them.  Instead of hanging around forever on the ``badges`` page, we should be able to click a button to remove them as applicants from the panel application.

Bjfitzpatrick:
Adds functionality to remove panelists


EliAndrewC:
+1

PR #72, opened 2017-01-07
Fixes #66

### LINKED ISSUE #66: Personalized schedules for panel staff
We have a personalized schedule for panelists showing the events to which they are assigned.  We should do the same thing for panel staff, showing a personalized schedule for every panel for which they are the point-of-contact.

Bjfitzpatrick:
Fixes #66


kitsuta:
Be sure to also actually describe what you're doing in the commit and PR description -- that way the history is much more readable.

panels/site_sections/panel_app_management.py
```diff
@@ -231,3 +231,26 @@ def everything(self, out, session):
                 ' / '.join(app.tech_needs_labels) + (' / ' if app.other_tech_needs else '') + app.other_tech_needs,
                 app.applied.strftime('%Y-%m-%d')
             ] + panelists)
+
+    def staff_schedule(self, session, attendee_id):
+        attendee = session.attendee(attendee_id)
+        events = defaultdict(lambda: defaultdict(lambda: (1, '')))
+        for pa in attendee.panel_applications:
```
kitsuta: Using super-short variables in loops like this is a bit of an anti-pattern (something done commonly that is not best practice). It's much easier to read the code if you use a meaningful variable name instead of a super-short acronym. In this case you can actually use `app` which is both descriptive and short!

panels/site_sections/panel_app_management.py
```diff
@@ -231,3 +231,26 @@ def everything(self, out, session):
...
+            if pa.event is not None:
+                for timeslot in pa.event.half_hours:
+                    rowspan = pa.event.duration if timeslot == pa.event.start_time else 0
+                    events[timeslot][pa.event.location_label] = (rowspan, pa.event.name)
+
+        schedule = []
+        locations = sorted(set(sum([list(locations) for locations in events.values()], [])))
```
kitsuta: This looks like something I would write when trying to figure out how the heck to get the data structure to be what I want it to be, which makes me think that it can probably be done better. @EliAndrewC would know -- data structures are my weak point.

Lusarien:
Ah gotcha, the fix was to add personal schedulea for an attendee

panels/site_sections/panel_app_management.py
```diff
@@ -231,3 +231,26 @@ def everything(self, out, session):
...
+        if events:
```
kitsuta: I don't see anywhere where events could be anything except the dict you defined above. And that dict also doesn't make any sense (this might be me). What is supposed to be happening here?

panels/site_sections/panel_app_management.py
```diff
@@ -231,3 +231,26 @@ def everything(self, out, session):
...
+            when = min(events)
```
kitsuta: This is a bit minor, but `events` implies a list of "things" but `when` is a measure of time. I'm not actually sure what kind of thing is supposed to be in the `events` dict from reading this code. Maybe a better name is `event_times`?

panels/templates/panel_app_management/assigned_to.html
```diff
@@ -6,6 +6,9 @@
 
 You can download a CSV spreadsheet of all of {{ attendee.first_name }}'s panels
 <a href="panels_by_poc?poc_id={{ attendee.id }}">here</a>.
+</br>
+You can view a schedule of all of {{ attendee.first_name }}'s panels
+<a href="staff_schedule?attendee_id={{ attendee.id }}">here</a>.
```
kitsuta: Another minor naming suggestion, but I was briefly confused by the fact that this says "staff" but the other link says "poc." I see from the issue that those are the same thing in this case, so it'd be easier for future readers to name this page `panel_poc_schedule` or something.

kitsuta (commented):
I'm a little confused by some of the things in here; I'm assuming you tested them as working. I don't see anything major that needs to be changed.

panels/site_sections/panel_app_management.py
Bjfitzpatrick: This is based off of https://github.com/magfest/panels/blob/master/panels/site_sections/schedule.py#L251
so Eli recommended doing it this way, I can change it if it's required though.

panels/site_sections/panel_app_management.py
Bjfitzpatrick: This is more or less to check for if there are events, if there aren't it throws an error if this isn't included.

panels/site_sections/panel_app_management.py
Bjfitzpatrick: Updated

panels/site_sections/panel_app_management.py
EliAndrewC: Specifically, calling ``min()`` on an empty sequence raises an exception.

panels/site_sections/panel_app_management.py
EliAndrewC: If this looks confusing at first glance, then maybe we should add a ``flatten`` function to ``utils.py``.  Python doesn't have a ``flatten`` function in its standard library, so the typical approach if you want to change ``[[5, 6], [7. 8]]`` into ``[5, 6, 7, 8]`` is to say ``sum([[5, 6,], [7, 8]], [])`` which is a bit ugly but gets the job done.

Bjfitzpatrick:
oops, mucked up one of these changes, do not merge give me a sec

Bjfitzpatrick:
Okay fixed it, back to working

panels/site_sections/panel_app_management.py
```diff
@@ -231,3 +231,26 @@ def everything(self, out, session):
                 ' / '.join(app.tech_needs_labels) + (' / ' if app.other_tech_needs else '') + app.other_tech_needs,
                 app.applied.strftime('%Y-%m-%d')
             ] + panelists)
+
+    def panel_poc_schedule(self, session, attendee_id):
+        attendee = session.attendee(attendee_id)
+        event_times = defaultdict(lambda: defaultdict(lambda: (1, '')))
+        for pa in attendee.panel_applications:
+            if pa.event is not None:
+                for timeslot in pa.event.half_hours:
```
EliAndrewC: I like @kitsuta's suggestion to change the ``pa`` variable name to ``app``, which is still a short enough variable name to not be confusing but is more descriptive - can you make that change?

Bjfitzpatrick: Done

EliAndrewC:
+1

PR #73, opened 2017-01-07
Fixes #57

### LINKED ISSUE #57: Point of contact dropdown isn't in alphabetical order
This is confusing for people and should be easy to fix.

Bjfitzpatrick:
Adds a sort to config.py to return a sorted list for the points of contact, I am not sure if sorting this list would cause problems elsewhere, any feedback would be great!

### DIFF
--- a/panels/config.py
+++ b/panels/config.py
@@ -11,7 +11,7 @@ class Config:
                 for a in session.query(AdminAccount)
                                 .options(joinedload(AdminAccount.attendee))
                                 .filter(AdminAccount.access.contains(str(c.PANEL_APPS)))
-            ],key = lambda tup: tup[1], reverse=False)
+            ], key = lambda tup: tup[1], reverse=False)
 
     @property
     def PANEL_ACCEPTED_EMAIL_APPROVED(self):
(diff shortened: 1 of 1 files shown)


panels/config.py
```diff
@@ -6,12 +6,12 @@ class Config:
...
     def PANEL_POC_OPTS(self):
         with Session() as session:
-            return [
+            return sorted([
                 (a.attendee.id, a.attendee.full_name)
                 for a in session.query(AdminAccount)
                                 .options(joinedload(AdminAccount.attendee))
                                 .filter(AdminAccount.access.contains(str(c.PANEL_APPS)))
-            ]
+            ],key = lambda y: y[1], reverse=False)
```
EliAndrewC: Can you rename this lambda parameter to be ``tup`` (for ``tuple``) instead of ``y``?  The latter will probably confuse people.

EliAndrewC:
> I am not sure if sorting this list would cause problems elsewhere, any feedback would be great!

Nope, shouldn't cause any problems.  I left a request for a variable name change but other than that this looks good to go.

EliAndrewC:
....except for pep8, which looks like it's failing.

PR #74, opened 2017-01-07
Fixes #67

### LINKED ISSUE #67: Point-of-contact CSV export should include application status
There's a CSV export of all panel applications for a given point of contact, but it doesn't say which are accepted, waitlisted, or declined.  We should add that as a column, possibly even the first column since it's pretty important.

Bjfitzpatrick:
Added extra column for labeling application status


kitsuta (approved):

EliAndrewC:
+1

PR #75, opened 2017-03-13
Fixes #2383: Adds static ident arg to each AutomatedEmail instance

RobRuana:
This pull request directly depends on [#2386](https://github.com/magfest/ubersystem/pull/2386).


kitsuta (approved):
+1

PR #78, opened 2017-04-28
Fixes ubersystem #2463: Replaces django's forloop.counter with jinja's loop.index

RobRuana:
Fixes https://github.com/magfest/ubersystem/issues/2463


binary1230:
+1 LGTM if that indeed fixes it

PR #80, opened 2017-05-05
Fix popup link on app management page.

kitsuta:
Fixes https://github.com/magfest/panels/issues/79.


kitsuta:
Merging this now since it prevents admins from using the live site.

PR #94, opened 2017-07-19
Allows more liberal movement between panel app status.

RobRuana:
Fixes magfest/ubersystem/issues/2683
Fixes magfest/ubersystem/issues/2684

Admins can now move a panels application from the cancelled
state back to accepted, waitlisted, or declined. Also changes
the cancel wording from "Cancel Application" to "Cancel Panel",
per #2683.


kitsuta (approved):
LGTM +1!

PR #96, opened 2017-07-25
Add several fields to panel application

kitsuta:
Fixes https://github.com/magfest/panels/issues/88. Also consolidates some copy-pasted fields into one included form.

### DIFF
--- a/panels/templates/panel_app_form.html
+++ b/panels/templates/panel_app_form.html
@@ -1,3 +1,29 @@
+<script type="text/javascript">
+    var showOrHideOtherPresentation = function () {
+        setVisible($.field('other_presentation').parents('.form-group'), $.val('presentation') === {{ c.OTHER }});
+    };
+    var showOrHideOtherLength = function() {
+        setVisible($('#other-length'), $.val('length') === {{ c.OTHER }});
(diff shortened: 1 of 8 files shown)


panels/templates/panel_app_form.html
```diff
@@ -0,0 +1,136 @@
...
+        <textarea class="form-control" name="description" rows="4">{{ app.description }}</textarea>
+    </div>
+</div>
+<div class="form-group">
+    <p class="help-block col-sm-9 col-sm-offset-3">
+        Will appear as a subheading on the schedule. 200 words max.
+    </p>
+</div>
+<div class="form-group" style="margin-bottom:0px">
+    <label class="col-sm-3 control-label">Expected Panel Length:</label>
```
kitsuta, line 59: Most of this is copy-pasted from the panel form, since we were repeating the same code in two places. The new fields start here and end at Technical Needs.

kitsuta:
The diffs are pretty ugly here, but most of the deletions/adds just represent stuff being moved.

RobRuana (approved):
Welp, looks better than what I had! My bad, +1!

PR #100, opened 2017-07-31
Improve data migration for panels to fix length display

kitsuta:
We also now display the panel length_text as long as it exists.


RobRuana (approved):
+1! Glad we were able to fix this up 😄

PR #101, opened 2017-07-31
Removes csrf checks from panel submissions.

RobRuana:
/panel_applications/index is now cached via nginx, so we can't dynamically generate a csrf token for the page. Fortunately we don't need to check for it in this case, because that page doesn't have any exploitable csrf vulnerabilities.


kitsuta (approved):
+1!

PR #102, opened 2017-08-01
Fix timespan printed in emails to panelists

kitsuta:
Panelists were being told that their panels would last twice as long as they actually will. This fixes that by specifying that the event length is counted in half-hour increments, not hour-increments like the timespan() function expects.


RobRuana (approved):
+1!

PR #104, opened 2017-08-03
Fixes #82, fixes #89, fixes #90: Adds panels application fields.

### LINKED ISSUE #82: Collect better Tech Ops info in the panel application form
We've found over the past few years that panelists are *really* bad at providing the necessary information about their technical needs.  For example, it's common for them to simply say they need "an adaptor" with no additional context.

We don't currently have a solid idea of what questions we need to ask, but we should probably have an entire section of the form devoted to their technical needs.  It can ask them a series of questions, such as "Are you bringing a laptop?" and then provide the subsequent questions as appropriate.

We'll need to ask Tech Ops for a set of information that would be useful to them and then extrapolate the list of questions that will wring that out of the applicants.

### LINKED ISSUE #89: Add new "Panel Categories" multiselect to both applications and events
The panel application form has a "Type of Panel" dropdown.  We'd like to replace that with a multiselect group of checkboxes with a set of predefined categories.

The panel folks will need to let us know what those categories should be.

One of the options should be "Other" which allows them to specify a freeform description.

Events should have this same set of checkboxes.  It should be required for panel applications but optional for events.

The panel folks would also like a report showing the overlaps of simultaneous events with the same categories.  For example, the panel department likes to avoid having two different "Game Show" events at the same time.  This should be allowed but simply be reported.

### LINKED ISSUE #90: Request extra information about each panelist
We currently ask for name, email, and phone number for each panelist.

The panels department has other information they'd sometimes like to collect (but which should probably be optional), such as social media info, what position that person has within the industry, etc.

The panels department will need to let us know exactly what information they want to collect before we can implement this.  Until that time, this issue is mostly a placeholder.

RobRuana:
Requires https://github.com/magfest/ubersystem/pull/2773

Fixes #82
Fixes #89
Fixes #90

### DIFF
--- a/alembic/versions/a3d71270256c_adds_fields_to_panel_applicant.py
+++ b/alembic/versions/a3d71270256c_adds_fields_to_panel_applicant.py
@@ -1,4 +1,4 @@
-"""Adds panel application fields
+"""Adds fields to panel_applicant
 
 Revision ID: a3d71270256c
 Revises: 416eb615ff1a
--- a/alembic/versions/a3d71270256c_adds_panel_application_fields.py
+++ b/alembic/versions/a3d71270256c_adds_panel_application_fields.py
@@ -1,0 +1,79 @@
+"""Adds panel application fields
+
+Revision ID: a3d71270256c
+Revises: 416eb615ff1a
+Create Date: 2017-08-03 15:42:52.406397
+
+"""
+
+
+# revision identifiers, used by Alembic.
+revision = 'a3d71270256c'
+down_revision = '416eb615ff1a'
+branch_labels = None
+depends_on = None
+
+from alembic import op
+import sqlalchemy as sa
+import sideboard
+
+
+
+try:
+    is_sqlite = op.get_context().dialect.name == 'sqlite'
+except:
+    is_sqlite = False
+
+if is_sqlite:
+    op.get_context().connection.execute('PRAGMA foreign_keys=ON;')
+    utcnow_server_default = "(datetime('now', 'utc'))"
+else:
+    utcnow_server_default = "timezone('utc', current_timestamp)"
+
+def sqlite_column_reflect_listener(inspector, table, column_info):
+    """Adds parenthesis around SQLite datetime defaults for utcnow."""
+    if column_info['default'] == "datetime('now', 'utc')":
+        column_info['default'] = utcnow_server_default
+
+sqlite_reflect_kwargs = {
+    'listeners': [('column_reflect', sqlite_column_reflect_listener)]
+}
+
+# ===========================================================================
+# HOWTO: Handle alter statements in SQLite
+#
+# def upgrade():
+#     if is_sqlite:
+#         with op.batch_alter_table('table_name', reflect_kwargs=sqlite_reflect_kwargs) as batch_op:
+#             batch_op.alter_column('column_name', type_=sa.Unicode(), server_default='', nullable=False)
+#     else:
+#         op.alter_column('table_name', 'column_name', type_=sa.Unicode(), server_default='', nullable=False)
+#
+# ===========================================================================
+
+
+def upgrade():
+    if is_sqlite:
+        with op.batch_alter_table('panel_applicant', reflect_kwargs=sqlite_reflect_kwargs) as batch_op:
+            batch_op.add_column(sa.Column('communication_pref', sa.Unicode(), server_default='', nullable=False))
+            batch_op.add_column(sa.Column('occupation', sa.Unicode(), server_default='', nullable=False))
+            batch_op.add_column(sa.Column('other_communication_pref', sa.Unicode(), server_default='', nullable=False))
+            batch_op.add_column(sa.Column('other_credentials', sa.Unicode(), server_default='', nullable=False))
+            batch_op.add_column(sa.Column('social_media', sideboard.lib.sa.JSON(), server_default='{}', nullable=False))
+            batch_op.add_column(sa.Column('website', sa.Unicode(), server_default='', nullable=False))
+    else:
+        op.add_column('panel_applicant', sa.Column('communication_pref', sa.Unicode(), server_default='', nullable=False))
+        op.add_column('panel_applicant', sa.Column('occupation', sa.Unicode(), server_default='', nullable=False))
+        op.add_column('panel_applicant', sa.Column('other_communication_pref', sa.Unicode(), server_default='', nullable=False))
+        op.add_column('panel_applicant', sa.Column('other_credentials', sa.Unicode(), server_default='', nullable=False))
+        op.add_column('panel_applicant', sa.Column('social_media', sideboard.lib.sa.JSON(), server_default='{}', nullable=False))
+        op.add_column('panel_applicant', sa.Column('website', sa.Unicode(), server_default='', nullable=False))
+
+
+def downgrade():
+    op.drop_column('panel_applicant', 'website')
+    op.drop_column('panel_applicant', 'social_media')
+    op.drop_column('panel_applicant', 'other_credentials')
+    op.drop_column('panel_applicant', 'other_communication_pref')
+    op.drop_column('panel_applicant', 'occupation')
+    op.drop_column('panel_applicant', 'communication_pref')
--- a/panels/configspec.ini
+++ b/panels/configspec.ini
@@ -16,6 +16,21 @@ panel_rooms = string_list(default=list("panels_1", "panels_2", "panels_3", "pane
 # deprecated "peglegs_email_signature" from the main repo.
 panels_email_signature = string(default=" - MAGFest Panels Team")
 
+# A list of social media fields collected from panelists.
+social_media = string_list(default=list("Facebook", "Twitter", "Instagram", "LinkedIn"))
+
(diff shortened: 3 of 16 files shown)


alembic/versions/a3d71270256c_adds_panel_application_fields.py
```diff
@@ -0,0 +1,79 @@
+"""Adds panel application fields
```
kitsuta: This is a really minor nitpick, but if I'm looking at the list of migration files, I'm not gonna know the difference between "add additional fields to panel app" and "adds panel application fields". Can we rename the file to something like "add even more fields to panel app"? If you can come up with something more descriptive that's even better, I just know that in six months I'm gonna have no idea which file is which. Feel free to rename the old migration if that helps!

panels/configspec.ini
```diff
@@ -16,6 +16,21 @@ panel_rooms = string_list(default=list("panels_1", "panels_2", "panels_3", "pane
 # deprecated "peglegs_email_signature" from the main repo.
 panels_email_signature = string(default=" - MAGFest Panels Team")
 
+# A list of social media fields collected from panelists.
+social_media = string_list(default=list("Facebook", "Twitter", "Instagram", "LinkedIn"))
```
kitsuta: It looks like this has some special relationship with the enums. If config options should match in certain ways, we should note that in a comment.

panels/models.py
```diff
@@ -10,6 +18,42 @@ def panel_applicants(self):
...
+    @property
+    def has_social_media(self):
+        return any(getattr(self, f) for f in self.social_media_fields.keys())
+
+    def __getattr__(self, name):
+        if name.endswith('_url'):
+            field_name = self.unqualify(name[:-4])
+            if field_name in self.social_media_fields:
+                attr = super(SocialMediaMixin, self).__getattr__(field_name)
+                attr = attr.strip('@#?=. ') if attr else ''
```
kitsuta, line 39: Minor cleanup suggestion: you check for `if attr` in the line after this, so you should be able to just move this under that block.

panels/templates/panel_app_form.html
```diff
@@ -9,6 +9,9 @@
             isNotAnHour = $.val('length') != {{ c.SIXTY_MIN }};
         setVisible($('#other-length'), isOther);
         $.field('length_text').prop('required', isOther);
+        if (!isOther) {
```
kitsuta, line 12: Bit of a warning: this means that if anyone goes to edit a panel application in Labs or West, the original data (the length entered by the panelist) will be removed from the panel app.

If you're okay with that, I'm fine with it -- your data migration looked really solid. Just wanted to double-check!

panels/templates/panel_applications/index.html
```diff
@@ -1,26 +1,30 @@
...
     };
     var showOrHideOtherPanelists = function () {
-        var count = $.val('other_panelists');
-        var $ops = $('#other-panelist-data tbody tr');
-        setVisible('#other-panelist-data', count);
+        var count = $.val('other_panelists'),
+            $template = $('#panelist_form_template > .panel'),
+            $ops = $('#panelists_container > .panel');
+        setVisible('#panelists_container', count);
+        console.log($template);
```
kitsuta: Is this a debug print-out?

panels/templates/panel_macros.html
```diff
@@ -0,0 +1,168 @@
...
+
+{% macro panelist_form(
+    panelist,
+    suffix='',
+    include_cellphone=True,
+    include_communication_pref=True,
+    is_readonly=False,
+    is_admin=False) -%}
+
+  {%- set is_class = panelist is class -%}
```
kitsuta, line 12: Where does `class` come from here? Is it a built-in variable? A brief Google search didn't turn up anything.

panels/models.py
```diff
@@ -10,6 +18,42 @@ def panel_applicants(self):
...
+        return cls.social_media_placeholders.get(name, '')
+
+    @property
+    def has_social_media(self):
+        return any(getattr(self, f) for f in self.social_media_fields.keys())
+
+    def __getattr__(self, name):
+        if name.endswith('_url'):
+            field_name = self.unqualify(name[:-4])
+            if field_name in self.social_media_fields:
```
kitsuta, line 37: Where does `social_media_fields` come from...? I see it referenced in multiple places but I don't see where it's defined. I'm sure I'm just missing something, but I figured I'd ask just in case.

kitsuta (approved):
This is... super fancy. :) I have a vague concern that the complexity will make it kind of difficult to edit the form in the future, but it looks like the business logic is pretty complicated so it probably can't be helped.

panels/models.py
kitsuta: Ah, I see it's all defined in the core plugin!

alembic/versions/a3d71270256c_adds_panel_application_fields.py
RobRuana: The names are totally arbitrary, but you're right – we should probably try to make them more meaningful. I'll go back and be more descriptive!

panels/configspec.ini
RobRuana: Yeah, I'll add a comment explaining the relationship. The `[social_media_urls]` and `[social_media_placeholders]` are nice to have, but not really necessary for the operation of the actual `SocialMediaMixin()` class.

panels/models.py
RobRuana: My concern was if somehow `attr` was somehow set to an all-whitespace string, like `"    "`.  It went through several different iterations, and I think I ended up with the worst version of the bunch ☹️ 

Could definitely use your cleanup suggestion.

panels/templates/panel_app_form.html
RobRuana: That's true! I didn't even think of that!

The thing I was trying to prevent is the following:
1. Someone chooses "Other"
2. Then they type 3 hours into the length
3. Then they change their mind and choose "2 hours"
4. Both fields get saved when they submit the form
5. Now the backend will display "3 hours" when it really should be "2 hours" 😦

RobRuana: I think all the data on labs/west is how it's supposed to be, so we should be safe. But I'll double check.

panels/templates/panel_applications/index.html
RobRuana: Yup! My bad!

panels/templates/panel_macros.html
RobRuana: It's defined in the core plugin.

panels/models.py
RobRuana: Ok, actually, that was correct as it was. If `attr` is `None` then we can't call `strip()` on it. And if `attr` consists entirely of these characters `@#?=. ` (including spaces), then we want to return an empty string.

So we do the first if statement to handle the case where `attr` is `None`, and the second if statement to handle the case where `attr` is stripped down to the empty string.

alembic/versions/a3d71270256c_adds_panel_application_fields.py
RobRuana: I'm going to update the descriptions to specify the actual table names we're modifying. I can't figure out how to be more descriptive without actually just listing the columns we're adding ☹️

RobRuana: Also, I'm going to remove the terms "additional" and "more" because they're redundant. Trying to make it concise and descriptive is actually quite difficult ☹️

panels/models.py
kitsuta: Ahh, I see! My bad!

alembic/versions/a3d71270256c_adds_fields_to_panel_applicant.py
```diff
@@ -0,0 +1,79 @@
+"""Adds fields to panel_applicant
```
kitsuta, line 1: +1 on the naming changes! I have a shot at remembering the difference now! :D

panels/configspec.ini
```diff
@@ -16,6 +16,30 @@ panel_rooms = string_list(default=list("panels_1", "panels_2", "panels_3", "pane
...
 
+# A list of social media fields collected from panelists.
+# The values in the list will be "fieldified" – converted to field names
+# (lowercased, spaces and special characters removed, CamelCase to
+# under_scored) and added as attributes on the PanelApplicant class.
+social_media = string_list(default=list("Facebook", "Twitter", "Instagram", "LinkedIn"))
+
+# These are used as templates to generate social media links, in case a
+# panelist provides a username instead of a direct link to their social media
+# page. The keys should match "fieldified" values in the "social_media" list.
```
kitsuta, line 27: In almost all other cases our key names are irrelevant, so it's super good that you added these notes. :)

panels/templates/panel_app_form.html
kitsuta: Yeah, I agree that we should blank out the field if the user goes from Other to another option. This is totally fine to do that as long as we're sure West and Labs' data are good. Worst case, we have it all in the tracking table, too.

kitsuta (approved):
+1!

PR #106, opened 2017-09-15
Fixes ability to uncheck checkgroups on various panel forms

RobRuana:
As it stands, you if any items in a check group are selected, you can't go back and uncheck them. So, if I create a panel application with "Projector" selected for "Technical Needs", and can't go back and edit that application and remove "Projector".

This pull request fixes that.


kitsuta (approved):
Haha, WHOOPS. Good catch!

PR #107, opened 2017-09-15
Fixes #86: Adds link to create new panel application to main panel application list

### LINKED ISSUE #86: Add a link to the Panel Application form on the admin side
After the panel application deadline has passed, logged-in admins are still able to fill it out in order to manually create new applications.  However, they often don't have easy access to that link and usually won't remember the specific URL they have to enter in order to get to it.

It would therefore be good to have a link at the top of the list of panel applications on the admin site which indicates that it can be followed to create new applications.

RobRuana:


kitsuta (approved):
+11111

PR #108, opened 2017-09-16
Add panel application for pre-approved panelists

kitsuta:
This is the second major refactor required for https://github.com/magfest/ubersystem/issues/2489 -- it adds a panel application form tailored for guests and other pre-approved and pre-existing panelists. It also factors out some panel processing functions to avoid code duplication.

### DIFF
--- a/panels/configspec.ini
+++ b/panels/configspec.ini
@@ -25,7 +25,7 @@ social_media = string_list(default=list("Facebook", "Twitter", "Instagram", "Lin
 # How many applications someone can submit as a confirmed PoC.
 # This limits how many applications someone can submit via the guest form.
 # Set this to 0 to remove limits from guest panel applications.
-app_limit = integer(default=0)
+app_limit = integer(default=3)
 
 # These are used as templates to generate social media links, in case a
 # panelist provides a username instead of a direct link to their social media
--- a/panels/configspec.ini
+++ b/panels/configspec.ini
@@ -22,6 +22,11 @@ panels_email_signature = string(default=" - MAGFest Panels Team")
 # under_scored) and added as attributes on the PanelApplicant class.
 social_media = string_list(default=list("Facebook", "Twitter", "Instagram", "LinkedIn"))
 
+# How many applications someone can submit as a confirmed PoC.
+# This limits how many applications someone can submit via the guest form.
+# Set this to 0 to remove limits from guest panel applications.
+app_limit = integer(default=0)
+
 # These are used as templates to generate social media links, in case a
 # panelist provides a username instead of a direct link to their social media
 # page. The keys should match "fieldified" values in the "social_media" list.
@@ -82,6 +87,7 @@ custom_av = string(default="Custom A/V Setup")
 
 [[presentation]]
 qa = string(default="Mostly just Question and Answer")
+group_discussion = string(default="Group Discussion with Moderator")
 basic = string(default="Basic presentation with some Q&A")
 advanced_av = string(default="Presentation requiring advanced A/V")
 workshop = string(default="Workshop or Tech/Interactive Demo")
--- a/panels/site_sections/panel_applications.py
+++ b/panels/site_sections/panel_applications.py
@@ -61,7 +61,7 @@ class Root:
             if not message:
                 message = process_panel_app(session, app, panelist, **params)
                 if not message:
-                    raise HTTPRedirect('guest?poc_id={}message={}', poc_id, 'Your panel application has been submitted')
+                    raise HTTPRedirect('index?message={}', 'Your panel application has been submitted')
 
         return {
             'app': app,
--- a/panels/site_sections/panel_applications.py
+++ b/panels/site_sections/panel_applications.py
@@ -17,7 +17,7 @@ def check_other_panelists(other_panelists):
             return '{} (for Other Panelist #{})'.format(message, i + 1)
 
 
-def check_extra_verifications(other_panelists=False, **params):
+def check_extra_verifications(**params):
     """
     Panelists submitting an application not associated with an attendee have some extra checkboxes they have
     to tick, so we validate them all here.
@@ -27,11 +27,11 @@ def check_extra_verifications(other_panelists=False, **params):
             c.EXPECTED_RESPONSE)
     elif 'verify_tos' not in params:
         return 'You must accept our Terms of Accommodation'
-    elif other_panelists and 'verify_poc' not in params:
+    elif 'other_panelists' in params and 'verify_poc' not in params:
         return 'You must agree to being the point of contact for your group'
 
 
-def other_panelists_from_params(app, **params):
+def compile_other_panelists_from_params(app, **params):
     # Turns form fields into a list of dicts of extra panelists on a panel application.
     other_panelists = []
     for i in range(1, int(params.get('other_panelists', 0)) + 1):
@@ -55,11 +55,12 @@ class Root:
         panelist = session.panel_applicant(panelist_params, checkgroups=PanelApplicant.all_checkgroups, restricted=True, ignore_csrf=True)
         panelist.application = app
         panelist.submitter = True
+        other_panelists = compile_other_panelists_from_params(app, **params)
 
         if cherrypy.request.method == 'POST':
-            message = check(panelist) or check_extra_verifications(int(params.get('other_panelists', 0)), **params)
+            message = check(panelist) or check_extra_verifications(**params)
             if not message:
-                message = process_panel_app(session, app, panelist, **params)
+                message = process_panel_app(session, app, panelist, other_panelists, **params)
                 if not message:
                     raise HTTPRedirect('index?message={}', 'Your panel application has been submitted')
 
@@ -67,7 +68,7 @@ class Root:
             'app': app,
             'message': message,
             'panelist': panelist,
-            'other_panelists': other_panelists_from_params(app, **params),
+            'other_panelists': other_panelists,
             'verify_tos': params.get('verify_tos'),
             'verify_poc': params.get('verify_poc'),
             'verify_waiting': params.get('verify_waiting'),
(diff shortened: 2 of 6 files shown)


panels/configspec.ini
```diff
@@ -82,6 +87,7 @@ custom_av = string(default="Custom A/V Setup")
 
 [[presentation]]
 qa = string(default="Mostly just Question and Answer")
+group_discussion = string(default="Group Discussion with Moderator")
```
kitsuta, line 90: This was added to default config to accommodate a new help block that asks panelists to specify who will moderate a panel. I figured this was probably common enough to include in the default config.

panels/configspec.ini
```diff
@@ -22,6 +22,11 @@ panels_email_signature = string(default=" - MAGFest Panels Team")
 # under_scored) and added as attributes on the PanelApplicant class.
 social_media = string_list(default=list("Facebook", "Twitter", "Instagram", "LinkedIn"))
 
+# How many applications someone can submit as a confirmed PoC.
+# This limits how many applications someone can submit via the guest form.
+# Set this to 0 to remove limits from guest panel applications.
+app_limit = integer(default=0)
```
kitsuta: Whoops, this is supposed to default to 3.

panels/site_sections/panel_applications.py
```diff
@@ -32,34 +55,72 @@ def index(self, session, message='', **params):
...
-                elif 'verify_tos' not in params:
-                    message = 'You must accept our Terms of Accomodation'
-                elif other_panelists and 'verify_poc' not in params:
-                    message = 'You must agree to being the point of contact for your group'
-                else:
-                    session.add_all([app, panelist] + other_panelists)
-                    raise HTTPRedirect('index?message={}', 'Your panel application has been submitted')
+                message = process_panel_app(session, app, panelist, **params)
+                if not message:
+                    raise HTTPRedirect('guest?poc_id={}message={}', poc_id, 'Your panel application has been submitted')
```
kitsuta: Wait, whoops, this redirect is wrong!

panels/site_sections/panel_applications.py
```diff
@@ -32,34 +55,72 @@ def index(self, session, message='', **params):
...
 
         return {
             'app': app,
             'message': message,
             'panelist': panelist,
-            'ops_count': len(other_panelists),
```
kitsuta, line 59: The `ops_count` is just replaced by a `other_panelists|length` since it's used only once in the template.

panels/site_sections/panel_applications.py
```diff
@@ -32,34 +55,72 @@ def index(self, session, message='', **params):
...
+            'verify_unavailable': params.get('verify_unavailable')
+        }
+
+
+def process_panel_app(session, app, panelist, **params):
+    """
+    Checks various parts of a new panel application, either submitted by guests or by non-attendees,
+    and then adds them to a session.
+    """
+    other_panelists = other_panelists_from_params(app, **params)
```
kitsuta: This all was a bit odd. I chose to do this as a way to get around a problem I had passing `other_panelists` as an argument to the `process_panel_app` function. The function would try to unpack the list, resulting in a "too many arguments" error. So rather than doing that, and rather than calling the `other_panelists_from_params` function several times, we only grab the other panelists when we actually need them -- when rendering the forms (in the return statements above) and when processing the app here.

panels/templates/panel_applications/guest.html
```diff
@@ -0,0 +1,81 @@
+{% extends "preregistration/preregbase.html" %}
+{% import 'panel_macros.html' as panel_macros %}
+{% block title %}Panel Application{% endblock %}
+{% block backlink %}{% endblock %}
+{% block content %}
+  {% include "panels_common.html" %}
+
+  {# This needs to be placed outside the <form> tag or we get validation errors #}
```
kitsuta: I really wanted to have the Other Panelists section self-contained, but when it was in the new template I made, it threw validation errors because the input tags were considered part of the form. :(

panels/site_sections/panel_applications.py
```diff
@@ -32,34 +55,72 @@ def index(self, session, message='', **params):
         panelist = session.panel_applicant(panelist_params, checkgroups=PanelApplicant.all_checkgroups, restricted=True, ignore_csrf=True)
         panelist.application = app
         panelist.submitter = True
-        other_panelists = []
-        for i in range(1, int(params.get('other_panelists', 0)) + 1):
-            applicant = {attr: params.get('{}_{}'.format(attr, i)) for attr in OTHER_PANELISTS_FIELDS}
-            other_panelists.append(PanelApplicant(application=app, **applicant))
```
RobRuana, line 38: _I'm not sure where to leave this comment, so I'm just putting it here..._

I'm not sure why the creation of `other_panelists` was moved to the end of this function. It seems like you end up calling `other_panelists_from_params()` twice. Once inside `process_panel_app()` and again inside the `return` statement. It seems like it's easier to just create them once right here, no?

panels/site_sections/panel_applications.py
RobRuana: Whoops, I mentioned this above, but don't you end up creating them twice?

panels/templates/other_panelists_form.html
```diff
@@ -0,0 +1,49 @@
...
+</div>
+
+<div id="panelists_container">
+  {% for op in other_panelists %}
+    <div class="panel panel-default">
+      <div class="panel-body">
+        <h2>Other Panelist #{{ loop.index }}</h2>
+        {{ panel_macros.panelist_form(
+            op,
+            suffix='_' + loop.index0|string,
```
RobRuana: This loop may need to start at index 1. So you may need to change this to `suffix='_' + loop.index|string,`.

panels/templates/panel_applications/guest.html
RobRuana: You could keep this with the Other Panelists section, and move it under the `<body>` element on document load.

```javascript
$(function () {
    $('#panelist_form_template').appendTo($('body'));
});
```

RobRuana (changes requested):
This looks mostly awesome. I've left a few comments. Some of which I'm not entirely sure about. I haven't actually tested this locally, but I am about to, so I may end up having more thoughts.

Really great work so far, though!

I'll post another review with my findings!

RobRuana (changes requested):
After loading this up locally, I get the following error when trying to submit a panel with multiple panelists:

```
Traceback (most recent call last):
  File "/home/vagrant/uber/sideboard/env/lib/python3.4/site-packages/cherrypy/_cprequest.py", line 670, in respond
    response.body = self.handler()
  File "/home/vagrant/uber/sideboard/env/lib/python3.4/site-packages/cherrypy/lib/encoding.py", line 221, in __call__
    self.body = self.oldhandler(*args, **kwargs)
  File "/home/vagrant/uber/sideboard/env/lib/python3.4/site-packages/cherrypy/_cpdispatch.py", line 60, in __call__
    return self.callable(*self.args, **self.kwargs)
  File "/home/vagrant/uber/sideboard/plugins/uber/uber/decorators.py", line 252, in with_timing
    return func(*args, **kwargs)
  File "/home/vagrant/uber/sideboard/plugins/uber/uber/decorators.py", line 267, in with_session
    retval = func(*args, session=session, **kwargs)
  File "/home/vagrant/uber/sideboard/plugins/uber/uber/decorators.py", line 388, in with_restrictions
    return func(*args, **kwargs)
  File "/home/vagrant/uber/sideboard/plugins/uber/uber/decorators.py", line 328, in with_rendering
    result = func(*args, **kwargs)
  File "/home/vagrant/uber/sideboard/plugins/panels/panels/site_sections/panel_applications.py", line 60, in index
    message = check(panelist) or check_extra_verifications(int(params.get('other_panelists', 0)), **params)
TypeError: check_extra_verifications() got multiple values for argument 'other_panelists'
```

I think my original review comments still stand.

panels/site_sections/panel_applications.py
kitsuta: Yup... I couldn't figure out a way around the error that I mentioned -- in fact, that same one you got. :\ This was a way to work around that error since I was really stuck on it, but clearly the fact that you ran into it means I need to actually figure it out.

RobRuana:
Oh! This is easy!

The first argument to `check_extra_verifications()` is `other_panelists`. `other_panelists` also exists as a keyword in `params`.

So the following statement:
```python
check_extra_verifications(int(params.get('other_panelists', 0)), **params)
```

really evaluates to something like this:
```python
check_extra_verifications(other_panelists=int(params.get('other_panelists', 0)), **{'other_panelists': 1})
```

So you really are passing `other_panelists` twice!

There's a couple ways to resolve this:
```python
# Ensure other_panelists is an int in the params dict
params['other_panelists'] = int(params.get('other_panelists', 0))
# No need to pass it separately, since it already exists in the params dict
check_extra_verifications(**params)
```

or:
```python
# Note the use of .pop() instead of .get()!
other_panelists = int(params.pop('other_panelists', 0))
check_extra_verifications(other_panelists, **params)
```

kitsuta:
Ooohhhhhhhhhhhhhhhhhhhhhhhhhhh thank goodness, I was super stuck but now
it's obvious in hindsight! I'll fix this stuff later tonight.

kitsuta:
OK, PR suggestions all implemented!

panels/templates/other_panelists_form.html
RobRuana: I don't think this loop is ever actually executed, so we haven't encountered this bug yet. But I still think the suffix needs to start at index 1 instead of index 0.

If you check the definition of `compile_other_panelists_from_params()`, you'll notice that the range over which it iterates starts at 1: `for i in range(1, int(params.get('other_panelists', 0)) + 1):`

RobRuana (changes requested):
One tiny change request. Otherwise 100% awesome!

panels/templates/other_panelists_form.html
kitsuta: Whoops, forgot to do that the first time!

RobRuana (approved):
+1!

PR #111, opened 2017-09-27
Customize panel application for guests

kitsuta:
Add an 'available' column so we can ask guests their for availability, not unavailability. Also removes the sixty minute checks, and allows a "return_to" parameter so guests can go back to their checklist when they're done.

### DIFF
--- a/panels/model_checks.py
+++ b/panels/model_checks.py
@@ -21,8 +21,7 @@ def overlapping_events(event, other_event_id=None):
 PanelApplication.required = [
     ('name', 'Panel Name'),
     ('description', 'Panel Description'),
-    ('length', 'Panel Length'),
-    ('unavailable', 'Your unavailability'),
+    ('length', 'Panel Length')
 ]
 PanelApplicant.required = [
     ('first_name', 'First Name'),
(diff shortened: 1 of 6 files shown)


panels/model_checks.py
```diff
@@ -45,36 +44,48 @@ def pa_phone(pa):
...
-    if localized_now() > c.PANEL_APP_DEADLINE and not c.HAS_PANEL_APPS_ACCESS:
+def app_deadline(app):
+    if localized_now() > c.PANEL_APP_DEADLINE and not c.HAS_PANEL_APPS_ACCESS and not app.poc_id:
         return 'We are now past the deadline and are no longer accepting panel applications'
 
 
 @validation.PanelApplication
-def specify_other_time(pa):
-    if pa.length == c.OTHER and not pa.length_text:
+def specify_other_time(app):
```
kitsuta, line 71: I was really sick of how confusing it was that the variable in "panel applicant" checks was named `pa` ... and the variable for "panel application" checks was also named `pa`. So I changed it.

panels/model_checks.py
```diff
@@ -45,36 +44,48 @@ def pa_phone(pa):
...
+    if app.length == c.OTHER and not app.length_text:
         return 'Please specify how long your panel will be.'
 
 
 @validation.PanelApplication
-def specify_nonstandard_time(pa):
-    if pa.length != c.SIXTY_MIN and not pa.length_reason:
+def specify_nonstandard_time(app):
+    if app.length != c.SIXTY_MIN and not app.length_reason and not app.poc_id:
```
kitsuta, line 78: This effectively bypasses this check for guests filling out the panel application.

panels/model_checks.py
```diff
@@ -45,36 +44,48 @@ def pa_phone(pa):
 
 
 @validation.PanelApplication
-def pa_other(pa):
-    if pa.presentation == c.OTHER and not pa.other_presentation:
+def unavailability(app):
```
kitsuta, line 47: While I could have simply changed the form to say "available" instead of "unavailable," I figured that will make things very hard for us when it comes to admins viewing panel information later on. So instead we check the "unavailable" field for non-guests and the "available" field for guests.

RobRuana (approved):
Haven't tried this out locally, but I don't see any problems!

Looks great!

PR #112, opened 2017-09-28
Fix validation always assuming 'other panelists' is filled out

kitsuta:
Not sure why, but 'other_panelists' was always in params, causing a false validation error for folks filling out the form. This moves the check back to the index function, where we can check the other_panelists that we generate from the params.


kitsuta:
This fixes a bug in production, so merging now!

RobRuana (commented):
+1, belatedly!

PR #114, opened 2017-09-29
Prevent Panels emails from being sent to guest badges

kitsuta:
The panels automated emails don't apply to the guest workflow, so we only send them if the receiver doesn't have a Guest badge.


RobRuana (approved):
+1!

PR #115, opened 2017-09-30
Fixes #85: Adds a "Show only panels rooms" checkbox to the schedule

### LINKED ISSUE #85: As an admin, I want to see a subset of rooms while editing the schedule
Panel admins only care about panel rooms.  It would make their lives a lot easier if they could toggle which rooms are being displayed while making edits to the schedule.

A simple version of this could be a "Show only panel rooms" checkbox.  A more complicated version of this could be giving admins full control over which rooms are shown.  Ideally we could do both, e.g. have some shortcuts ("Music Rooms", "Tabletop Rooms", "Panel Rooms") but also offer full control.

RobRuana:
Also cleans up and unifies the schedule styling.

screen shot 2017-09-30 at 6 59 11 pm

screen shot 2017-09-30 at 6 59 31 pm

### DIFF
--- a/panels/site_sections/schedule.py
+++ b/panels/site_sections/schedule.py
@@ -11,7 +11,6 @@ class Root:
         else:
             raise HTTPRedirect("internal")
 
-    @cached
     def internal(self, session, message=''):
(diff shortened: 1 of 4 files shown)


kitsuta (approved):
Neat!

PR #116, opened 2017-10-03
Fixes schedule layout and scrolling in internet explorer and firefox

RobRuana:
Fixes some issue with scrolling and layout in internet explorer and firefox. This may be easiest to review by loading it up locally; but please let me know if you see any glaring errors in the html/css/js.

### DIFF
--- a/panels/site_sections/schedule.py
+++ b/panels/site_sections/schedule.py
@@ -11,7 +11,6 @@ class Root:
         else:
             raise HTTPRedirect("internal")
 
-    @cached
     def internal(self, session, message=''):
         if c.HIDE_SCHEDULE and not AdminAccount.access_set() and not cherrypy.session.get('staffer_id'):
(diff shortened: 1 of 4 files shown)


panels/site_sections/schedule.py
```diff
@@ -11,7 +11,6 @@ def index(self, session, message=''):
         else:
             raise HTTPRedirect("internal")
 
-    @cached
```
kitsuta: You gotta re-cache this!

RobRuana: I can’t figure out why it was cached in the first place! But I’ll add the cache back 😅

kitsuta: I think it's cached because people would view it a lot on-site, hitting our servers real bad. There's probably better ways to handle that than caching it, but we've been bitten many times by the "oh we'll remember to fix this server-killing problem before we go onsite" pattern.

kitsuta (approved):
+1!

PR #117, opened 2017-10-03
Fixes schedule layout on IE (for real)

RobRuana:
I actually have access to a windows box, so I can test things this time around.


RobRuana:
Merging now, because the schedule is broke on production and I've been able to test this on windows locally.

PR #120, opened 2017-10-05
Fixes #61: Makes Attendee.panel_applicants a list instead of a one-to-one

### LINKED ISSUE #61: Panel applicant backref should be a list
Many different ``PanelApplicant`` entries can link to the same ``Attendee`` but the relationship is defined as one-to-one.  We should fix that, since it prevents us from e.g. easily iterating over all applicants for a given attendee.

RobRuana:


panels/models.py
```diff
@@ -57,7 +57,7 @@ def __getattr__(self, name):
 @Session.model_mixin
 class Attendee:
     assigned_panelists = relationship('AssignedPanelist', backref='attendee')
-    panel_applicant = relationship('PanelApplicant', backref='attendee', uselist=False)
+    panel_applicants = relationship('PanelApplicant', backref='attendee')
```
RobRuana, line 60: Surprisingly, this property was not referenced anywhere, in any of our plugins! So this is the only line that needed to change 😄

kitsuta (approved):
I guess this is because someone can apply for a panel multiple times in a row, creating many panel applicants for one person?

RobRuana:
Yeah, our data model is not perfect. If a single attendee is conducting multiple panels, there will be an associated `PanelApplicant` for each panel.

PR #121, opened 2017-10-05
Fixes #103: Adds list of panel applications to attendee form

### LINKED ISSUE #103: As an admin, I'd like to look at an Attendee record and know what panels apps they are associated with
if looking at the /uber/registration/form page, it would be cool if someone is associated with a panel app if it told you.  something like this;

image

situation that just happened:  Someone got an email saying "Hey congrats you're a panelist, please fill in your new placeholder badge" and they didn't know they had been named as a co-presenter of the panel.  From the email and attendee record, Gerardo wasn't able to easily figure out what panel they were associated with, and we could only get it by looking at the global feed of database changes, which is very user unfriendly.

I'd say low priority, but nice to have.

might need https://github.com/magfest/panels/issues/61 for this

RobRuana:
This depends on both:
* https://github.com/magfest/ubersystem/pull/2884
* https://github.com/magfest/panels/pull/120

### DIFF
--- a/panels/templates/registration/form.html
+++ b/panels/templates/registration/form.html
@@ -1,6 +1,6 @@
-{% extends "registration/baseform.html" %}
-
+{% extends "registration/form.html" %}
 {% block pre_regform -%}
+  {{ super() }}
   {% if attendee.panel_applicants -%}
     <div class="form-group">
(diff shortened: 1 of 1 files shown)


kitsuta (approved):
LGTM!

PR #122, opened 2017-10-09
Fixes #118: Moves validation before model objects are added to the session

### LINKED ISSUE #118: Bad panels application validation logic
The following line adds the newly created model objects to the database session (note that this happens inside the `process_panel_app()` method): https://github.com/magfest/panels/blob/f6f87ac286851e5d76d76edbb01535364a2b5a24/panels/site_sections/panel_applications.py#L127

However, one additional validation is performed *after* the items are added to the session (note that this happens immediately after the call to `process_panel_app()`): https://github.com/magfest/panels/blob/f6f87ac286851e5d76d76edbb01535364a2b5a24/panels/site_sections/panel_applications.py#L64

This means that the panel application could be added to the session and saved to the database while in an invalid state. The above validation should be moved to *before* the call to `process_panel_app()`.

### LINKED ISSUE #119: Intermittent email validation failures for Other Panelists
This is a tricky one, because we haven't been able to reproduce it at all. There have been two separate reports, both of which came in during the afternoon and evening of October 2, 2017. Same issue described in both reports. While attempting to submit a panels application with "Other Panelists", an validation error is reported for the "Other Panelist", even though in both cases a valid email was provided.

There has been speculation that this was caused by a bad copy/paste that somehow included invalid (non-printable) characters.

While investigating this issue, I ran into another (potentially related?) issue: https://github.com/magfest/panels/issues/118

I don't know if these issues are related, but maybe?

RobRuana:
The main purpose of this pull request was to fix #118. However, I believe this may also fix #119.

### DIFF
--- a/panels/templates/other_panelists_form.html
+++ b/panels/templates/other_panelists_form.html
@@ -52,7 +52,7 @@
         <h2>Other Panelist #{{ loop.index }}</h2>
         {{ panel_macros.panelist_form(
             op,
-            suffix='_' + loop.index1|string,
+            suffix='_' ~ loop.index,
             require_cellphone=False,
             require_communication_pref=False) }}
       </div>
--- a/panels/site_sections/panel_applications.py
+++ b/panels/site_sections/panel_applications.py
@@ -59,12 +59,12 @@ class Root:
 
         if cherrypy.request.method == 'POST':
             message = check(panelist) or check_extra_verifications(**params)
+            if not message and other_panelists and 'verify_poc' not in params:
+                message = 'You must agree to being the point of contact for your group'
             if not message:
                 message = process_panel_app(session, app, panelist, other_panelists, **params)
-                if not message and other_panelists and 'verify_poc' not in params:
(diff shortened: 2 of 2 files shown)


panels/templates/other_panelists_form.html
```diff
@@ -52,7 +52,7 @@
         <h2>Other Panelist #{{ loop.index }}</h2>
         {{ panel_macros.panelist_form(
             op,
-            suffix='_' + loop.index1|string,
```
RobRuana, line 55: This is the change that I believe fixes #119. If you submit additional "Other Panelists" with bad data, when the form was re-rendered after the POST, all of the "Other Panelists" form fields would have the wrong name. They'd all be like `first_name_` without the integer suffix. That will cause every subsequent submission of the form to fail!

kitsuta: Oh no! What a tricky bug, nice find!

kitsuta (approved):
+1!

PR #124, opened 2017-10-19
Allow admins to associate events with panel applications more seamlessly

kitsuta:
Right now, admins must create an event, then go to a totally different page to associate a panel application with that event. This seems silly, because it forces the admin to work in two separate places and to manually copy over panel names and descriptions. Now admins can associate events with panel applications while editing an event; additionally, if they associate a new event with a panel app, they can skip entering in name/desc. Any panelists associated with the panel app also get added to the event.

### DIFF
--- a/panels/models.py
+++ b/panels/models.py
@@ -99,7 +99,10 @@ class AssignedPanelist(MagModel):
     event_id    = Column(UUID, ForeignKey('event.id', ondelete='cascade'))
 
     def __repr__(self):
-        return '<{self.attendee.full_name} panelisting {self.event.name}>'.format(self=self)
+        if self.attendee:
+            return '<{self.attendee.full_name} panelisting {self.event.name}>'.format(self=self)
+        else:
+            return super(AssignedPanelist, self).__repr__()
 
 
 class PanelApplication(MagModel):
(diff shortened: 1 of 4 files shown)


panels/models.py
```diff
@@ -99,7 +99,10 @@ class AssignedPanelist(MagModel):
     event_id    = Column(UUID, ForeignKey('event.id', ondelete='cascade'))
 
     def __repr__(self):
-        return '<{self.attendee.full_name} panelisting {self.event.name}>'.format(self=self)
+        if self.attendee:
+            return '<{self.attendee.full_name} panelisting {self.event.name}>'.format(self=self)
+        else:
+            return super(AssignedPanelist, self).__repr__()
```
kitsuta, line 105: I had to do this to prevent `__repr__` errors when creating a new instance of this model, since the relationship doesn't have a chance to exist right when we create a new AssignedPanelist object.

panels/site_sections/schedule.py
```diff
@@ -173,6 +173,21 @@ def form(self, session, message='', panelists=(), **params):
...
+            # Associate a panel app with this event, and if the event is new, use the panel app's name and title
+            if 'panel_id' in params:
+                add_panel = session.panel_application(id=params['panel_id'])
+                add_panel.event_id = event.id
+                session.add(add_panel)
+                if event.is_new:
+                    event.name = add_panel.name
+                    event.description = add_panel.description
+                    for pa in add_panel.applicants:
+                        if pa.attendee_id:
```
kitsuta, line 186: For practical reasons, panelists without associated attendees cannot be associated with an event.

RobRuana (approved):
Looks really great, +1!

PR #126, opened 2017-10-23
Fixes case where someone is trying to create an event without an associated panel

RobRuana:
Emergency fix, cause someone is waiting on it! 😄


kitsuta (commented):
Oops! Belated +1!

PR #128, opened 2017-11-10
Fixes #2903: Adds show only music checkbox, and text based filtering of schedule

RobRuana:


RobRuana:
# Text Filter in Action
screen shot 2017-11-10 at 6 21 46 pm

panels/config.py
```diff
@@ -40,13 +40,29 @@ def PANEL_SCHEDULED_EMAIL_APPROVED(self):
 c.EVENT_BOOKED = {'colspan': 0}
 c.EVENT_OPEN   = {'colspan': 1}
 
-invalid_panel_rooms = [room for room in c.PANEL_ROOMS if not getattr(c, room.upper(), None)]
+nesteddict = lambda: defaultdict(nesteddict)
 
-for room in invalid_panel_rooms:
+
+def _make_room_trie(rooms):
```
kitsuta, line 46: What's a trie?

kitsuta (approved):
LGTM!

panels/config.py
RobRuana: Oh, this is exciting! I get to talk about one of the most unfortunately named data structures in computer science, the [Trie](https://en.wikipedia.org/wiki/Trie)!

I'm not going to repeat the whole wikipedia article here, but to summarize, a trie is a tree-based data structure that is perfect for this kind of text-filtering feature. Each node in the tree maps to a single character. As you descend through the branches of the tree, the path you take represents the string you're building.

In this particular implementation, I've cached each matching room location with each node in the tree. As you descend the tree, you have instant access to all the matching room locations. This tradeoff gives us much faster searching, but requires additional memory to store the duplicate room locations at each node. Since the additional memory is not prohibitive, it's a worthwhile tradeoff!

# Room Trie Example
```
>>> from panels.config import _make_room_trie
>>> from pprint import pprint
>>> pprint(_make_room_trie([(1111, 'Mages'), (2222, 'Magic'), (3333, 'Moo')]))
{'__rooms__': {1111: 0,
               2222: 1,
               3333: 2},
 'm': {'__rooms__': {1111: 0,
                     2222: 1,
                     3333: 2},
       'a': {'__rooms__': {1111: 0,
                           2222: 1},
             'g': {'__rooms__': {1111: 0,
                                 2222: 1},
                   'e': {'__rooms__': {1111: 0},
                         's': {'__rooms__': {1111: 0}}},
                   'i': {'__rooms__': {2222: 1},
                         'c': {'__rooms__': {2222: 1}}}}},
       'o': {'__rooms__': {3333: 2},
             'o': {'__rooms__': {3333: 2}}}}}
```

kitsuta: oh wow, that's super cool!

PR #130, opened 2017-11-17
Fixes schedule layout so horizontal scrollbar is visible

RobRuana:
This is basically reverting the changes from here: https://github.com/magfest/panels/pull/127. But I added a delay function to the scroll event handler to help scrolling performance a little bit.


kitsuta (commented):
+1 (belatedly)

PR #132, opened 2017-11-29
Relevant to #2491, #2627, #2866: Adds admin side of attractions feature

RobRuana:
Required by https://github.com/magfest/ubersystem/pull/2990


panels/models/attraction.py
```diff
@@ -0,0 +1,398 @@
...
+        (900, '15 minutes before'),
+        (1800, '30 minutes before'),
+        (3600, '1 hour before'),
+        (7200, '2 hours before'),
+        (86400, '1 day before')]
+
+    name = Column(UnicodeText, unique=True)
+    description = Column(UnicodeText)
+    notifications = Column(JSON, default=[], server_default='[]')
+    required_checkin = Column(Integer, default=0)  # In seconds
```
kitsuta, line 100: Is there any specific reason you didn't make this a Choice column with REQUIRED_CHECKIN_OPTS?

panels/models/attraction.py
```diff
@@ -0,0 +1,398 @@
...
+    @property
+    def used_location_opts(self):
+        locs = set(e.location for e in self.events)
+        sorted_locs = sorted(locs, key=lambda l: c.EVENT_LOCATIONS[l])
+        return [(l, c.EVENT_LOCATIONS[l]) for l in sorted_locs]
+
+    @property
+    def unused_location_opts(self):
+        locs = set(e.location for e in self.events)
+        return [(l, s) for l, s in c.EVENT_LOCATION_OPTS if l not in locs]
```
kitsuta, line 149: It's a bit strange that the used location opts are sorted, but the unused ones are not.

panels/models/attraction.py
```diff
@@ -0,0 +1,398 @@
...
+
+    __mapper_args__ = {'confirm_deleted_rows': False}
+    __table_args__ = (UniqueConstraint('attraction_event_id', 'attendee_id'),)
+
+
+Attraction.required = [('name', 'Name'), ('description', 'Description')]
+AttractionFeature.required = [('name', 'Name'), ('description', 'Description')]
+
+
+@validation.AttractionEvent
```
kitsuta, line 395: I'd say this should be moved into a model_checks.py file but I think you already did that!

panels/site_sections/attractions.py
```diff
@@ -0,0 +1,114 @@
...
+@check_post_con
+class Root:
+
+    def index(self, session, **params):
+        attractions = session.query(Attraction).options(
+            subqueryload(Attraction.features)).order_by(Attraction.name).all()
+        return {'attractions': attractions}
+
+    def features(self, session, id=None, **params):
+        attraction = _model_for_id(
```
kitsuta, line 47: What was the idea behind using _model_for_id instead of session.attraction(id=id)? Does the latter not allow subqueryloads?

panels/site_sections/attractions.py
```diff
@@ -0,0 +1,114 @@
...
+            session, Attraction, id,
+            subqueryload(Attraction.features)
+                .subqueryload(AttractionFeature.events)
+                    .subqueryload(AttractionEvent.attendees))
+
+        if not attraction:
+            raise HTTPRedirect('index')
```
kitsuta, line 54: This should probably have a message explaining the redirect. 😅

panels/site_sections/attractions.py
```diff
@@ -0,0 +1,114 @@
...
+        if not attendee:
+            return {'error': 'Unrecognized badge number: {}'.format(badge_num)}
+
+        event = _model_for_id(session, AttractionEvent, id)
+        if not event:
+            return {'error': 'Unrecognized event id: {}'.format(id)}
+
+        old_remaining_slots = event.remaining_slots
+
+        if event not in attendee.attraction_events:
```
kitsuta, line 89: This should probably be a series of model_checks -- on the other hand, model_checks doesn't support "only do these checks if a prerequisite is met" very well, so maybe we want to add that functionality to model_checks first, somehow (even if we make it only support "only do these checks if the value has changed" and "only do these checks for new models")

panels/site_sections/attractions_admin.py
```diff
@@ -0,0 +1,330 @@
...
+        return {
+            'filtered': filtered,
+            'message': message,
+            'attractions': attractions
+        }
+
+    def form(self, session, message='', **params):
+        attraction_id = params.get('id')
+        if not attraction_id or attraction_id == 'None':
+            raise HTTPRedirect('index')
```
kitsuta, line 30: Any reason not to just redirect to `new`..?

panels/site_sections/attractions_admin.py
```diff
@@ -0,0 +1,330 @@
...
+                session.add(attraction)
+                raise HTTPRedirect(
+                    'form?id={}&message={}',
+                    attraction.id,
+                    '{} updated successfully'.format(attraction.name))
+        else:
+            attraction = session.query(Attraction) \
+                .filter_by(id=attraction_id) \
+                .options(
+                    subqueryload(Attraction.department),
```
kitsuta, line 54: This is just to double-check, but if the attraction doesn't have a department set, this subqueryload won't cause any major problems, right?

panels/site_sections/attractions_admin.py
```diff
@@ -0,0 +1,330 @@
...
+                    "You cannot delete an attraction that you don't own")
+
+            session.delete(attraction)
+            raise HTTPRedirect(
+                'index?message={}',
+                'The {} attraction was deleted'.format(attraction.name))
+
+        raise HTTPRedirect('form?id={}', id)
+
+    def feature(self, session, attraction_id=None, message='', **params):
```
kitsuta, line 113: This is an unusual request, but I STILL have trouble keeping track of what's an attraction, what's a feature, and what's an event. At some point can you add a docstring that explains/gives an example, even if only at the top of this file?

Thinking about it, I wonder if we can/should at some point change `event` to `timeslot`. That may help a little.

kitsuta (commented):
I THINK I got through all of this... +1, of course. What a massive feature!

panels/models/attraction.py
RobRuana: Because I wanted this column to be able to store an arbitrary number of seconds. `REQUIRED_CHECKIN_OPTS` probably should have been declared somewhere else, like maybe directly in the HTML template? (I think that's the only place it's used)

panels/models/attraction.py
RobRuana: Meh, I didn't need `unused_location_opts` to be sorted ¯\\_(ツ)_/¯

panels/site_sections/attractions.py
RobRuana: Two reasons:
* You guessed the first, no subqueryloads!
* `session.attraction(id=id)` uses `query.one()`, which will throw an exception if no rows are found. You see a lot of those errors in our logs. I wanted to be more forgiving, so if someone enters an unknown id, I didn't want the app to blow up

panels/site_sections/attractions_admin.py
RobRuana: Should be okay

panels/site_sections/attractions_admin.py
RobRuana: Yeah, there's lots of different types of events throughout the system. I tried to use `attraction_event` where there might be ambiguity, and just `event` where the meaning would be clear through context.

Naming is one of the hardest things in computer science!

PR #133, opened 2017-12-03
Prevents guests from receiving panels emails

RobRuana:
The automated panels emails were checking for `PanelApplication.poc.badge_type != c.GUEST_BADGE`. The problem with that code is that the `poc` refers to the _staff_ point of contact – usually someone from the office or guests department, who will _never_ have a guest badge.

The code _should_ have been checking for `PanelApplication.submitter.attendee.badge_type != c.GUEST_BADGE`. The `submitter` is the guest – the person that actually submitted the panel application.

*NOTE*: This pull request also includes a small refactor of `panels_common.html`. It looks like a lot, but it's really just code copy/pasted into `schedule_common.html`, so the schedule filters don't show up on every panels page.


panels/automated_emails.py
```diff
@@ -1,16 +1,27 @@
...
-        AutomatedEmail.__init__(self, PanelApplication, subject, template,
-                                lambda app: (not app.poc_id or app.poc.badge_type != c.GUEST_BADGE) and filter(app),
-                                ident, sender=c.PANELS_EMAIL, **kwargs)
+        AutomatedEmail.__init__(
+            self,
+            PanelApplication,
+            subject,
+            template,
+            lambda app: filter(app) and (
+                not app.submitter or
```
kitsuta, line 19: This looks like it would prevent ANY panels emails from going out, since apps always have submitters?

The reason for `not app.poc_id` in the original filter is, for the guests' version of the panel application, we specifically set the POC to the submitter of the app as soon as the app is submitted (https://github.com/magfest/panels/blob/master/panels/site_sections/panel_applications.py#L89). If the POC should be the staff point of contact then we need to undo that as well, I guess. :( In any case, these two lines definitely seem like they'd break panels emails.

RobRuana: The poc is the _staff_ poc, so I think the logic in [this line](https://github.com/magfest/panels/blob/master/panels/site_sections/panel_applications.py#L89) is incorrect. 

It took me awhile to figure out that poc was referring to _staff_ poc! It really should have been named `staff_poc_id`, but as we all know, naming things is the hardest task in computer science!

kitsuta: Yeah, I agree that line is probably incorrect. Bleh. :(

Do we know if the system is willing to send panelists emails with this change? I can dig into it if we want, though I expect all the emails that would have been sent, have been sent.

RobRuana: I believe so, because panels admins assign themselves as poc, right?

kitsuta: I don't actually know, but I guess it's okay because even if the first two checks fail, the third one would succeed for all non-guest panel applicants. In that case, I think these two lines are unnecessary, but they don't cause any problems.

RobRuana: Oh, oh, right. I misinterpreted your first comment. Yeah, ok, here's why those lines exist. They are not actually relevant to the filter itself, but they are in place to prevent `AttributeError: 'NoneType' object has no attribute` errors.

It's entirely possible that a bug could be introduced in the program that causes `app.submitter` to be `None`, even though we don't ever want that to be the case. Let's look at the definition of `app.submitter`:
```python
    @property
    def submitter(self):
        for a in self.applicants:
            if a.submitter:
                return a
        return None
```

If, for some reason, none of the `applicants` is marked as the submitter, then `app.submitter` will be `None`. In that case, the following line will throw an exception:
```python
app.submitter.attendee.badge_type != c.GUEST_BADGE
```

Saying `not app.submitter or` prevents that error.

Ok, so lets say `app.submitter` has a valid value. `app.submitter.attendee` could ALSO be `None`! Let's look at the definition of `app.submitter.attendee_id`:
```python
attendee_id = Column(UUID, ForeignKey('attendee.id', ondelete='cascade'), nullable=True)
```

That can definitely be `None` because `nullable=True`. In fact, it _will_ be `None` until we've assigned an `Attendee` to the `PanelApplicant`. So again, the following line will throw an exception:
```python
app.submitter.attendee.badge_type != c.GUEST_BADGE
```

So before we execute that line, we have to do another `None` check like this: `not app.submitter.attendee_id or`. Put that altogether and you have the filter as it was written:

```python
lambda app: filter(app) and (
    not app.submitter or
    not app.submitter.attendee_id or
    app.submitter.attendee.badge_type != c.GUEST_BADGE)
```

I hope that makes sense!

😃

kitsuta: Ohhh I see, yes that makes sense. Was there any particular reason you didn't just check for truthiness and used `and`? So `app.submitter and app.submitter.attendee_id`. After all, the email templates access things like `app.submitter.first_name`, so if we try to send an email without a submitter, that would cause errors, right?

RobRuana: You're absolutely right! But I don't think we'd even get to the point where the email templates are rendered, because `app.email` will return `None`, and our email code will disregard it.

I based the `not/or` construction on what was there before without really thinking about it:
```
lambda app: (not app.poc_id or app.poc.badge_type != c.GUEST_BADGE) and filter(app)
```

PR #134, opened 2017-12-09
Improves panel automated email performance

RobRuana:


kitsuta (commented):
+1 (belatedly) and changed the title.

I guess now that I think about it, I could have named them all 'add dynamic property' and that would have worked too... oh well.

PR #137, opened 2017-12-19
Allows attraction checkin page to be used by attendee with the same permissions as a reg station

RobRuana:
I realized that attractions might have attendees working the checkin booth that might not have full admin privileges, like we do at reg. This allows someone with more limited access to still use the checkin page.

Also makes the URL a little nicer:
https://localhost:4443/uber/attractions_admin/checkin?id=escape-room
instead of:
https://localhost:4443/uber/attractions_admin/checkin?id=ff5828b7-1e48-4fd4-91ef-a38e4c95a1e1

### DIFF
--- a/panels/site_sections/attractions_admin.py
+++ b/panels/site_sections/attractions_admin.py
@@ -93,17 +93,6 @@ class Root:
             'message': message,
         }
 
-    def checkin(self, session, message='', **params):
-        attraction_id = params.get('id')
-        if not attraction_id or attraction_id == 'None':
-            raise HTTPRedirect('index')
-
-        attraction = session.query(Attraction) \
-            .filter_by(id=attraction_id) \
-            .order_by(Attraction.id).one()
-
-        return {'attraction': attraction, 'message': message}
-
     @csrf_protected
     def delete(self, session, id, message=''):
         if cherrypy.request.method == 'POST':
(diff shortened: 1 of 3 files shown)


RobRuana:
Also, when editing a time gap between events, it will only push other events in that same day forward.

panels/site_sections/attractions_admin.py
```diff
@@ -324,6 +314,42 @@ def delete_event(self, session, id):
...
+                message = "You cannot cancel a signup for an attraction you don't own"
+            elif signup.checkin_time:
+                message = "You cannot cancel a signup that has already checked in"
+            else:
+                session.delete(signup)
+                session.commit()
+        if message:
+            return {'error': message}
+
+    @only_renderable(c.STUFF, c.PEOPLE, c.REG_AT_CON)
```
kitsuta: Maybe we should just rename the REG_AT_CON permission to "volunteer" 😅  That's out of scope for this PR, of course.

panels/site_sections/attractions_admin.py
```diff
@@ -324,6 +314,42 @@ def delete_event(self, session, id):
...
+    def checkin(self, session, message='', **params):
+        id = params.get('id')
+        if not id:
+            raise HTTPRedirect('index')
+
+        try:
+            uuid.UUID(id)
+            filters = [Attraction.id == id]
+        except Exception:
+            filters = [Attraction.slug.startswith(sluggify(id))]
```
kitsuta, line 344: Oh man, this is a SUPER useful feature. We don't have to do it for this PR, but we should totally make this into a decorator. Imagine if any model with a `slug` column could have pretty URLs instead of ID-based URLs!

kitsuta (approved):
LGTM!

...GO TO BED

panels/site_sections/attractions_admin.py
RobRuana: Yes, yes, yes! Nice URLs everywhere!

PR #138, opened 2017-12-20
Attendees can sign up for attractions without badge numbers

EliAndrewC:
As discussed on the code@magfest.org mailing list earlier this week, this allows people to sign up for attractions before the event starts by entering their name/email/zipcode.

We originally talked about making this configurable on a per-feature basis, but it looks like Rob already added the ability for admins to toggle whether or not their attraction is ready for signups, which I believe handles that use case.

I tried a few different ways to add this, and I eventually settled on adding a "I don't have my badge yet" button to the signup modal:
newfooterbutton

which opens a different modal form:
newform

I spent some time trying to dynamically swap out the contents of the existing modal, but it was difficult to make the styling work due to the different sizes of the respective forms.  This approach works and doesn't involve any serious copy/paste.

I don't own a smartphone, but I tested this out in Google Chrome using their "Device Toolbar" in the Developer Tools.  The "I don't have my badge yet" button is a little big due to being wordy (happy to take suggestions for a less wordy name) but other than that everything looks okay to me.

### DIFF
--- a/panels/site_sections/attractions.py
+++ b/panels/site_sections/attractions.py
@@ -20,12 +20,10 @@ def _attendee_for_badge_num(session, badge_num, options=None):
 
 
 def _attendee_for_info(session, first_name, last_name, email, zip_code):
-    return session.query(Attendee).filter(
-        func.lower(Attendee.first_name) == first_name.lower(),
-        func.lower(Attendee.last_name) == last_name.lower(),
-        Attendee.normalized_email == Attendee.normalize_email(email),
-        Attendee.zip_code == zip_code
-    ).first()
+    try:
+        return session.lookup_attendee(first_name, last_name, email, zip_code)
+    except:
+        return None
 
 
 def _model_for_id(session, model, id, options=None, filters=[]):
--- a/panels/site_sections/attractions.py
+++ b/panels/site_sections/attractions.py
@@ -23,7 +23,7 @@ def _attendee_for_info(session, first_name, last_name, email, zip_code):
     return session.query(Attendee).filter(
         func.lower(Attendee.first_name) == first_name.lower(),
         func.lower(Attendee.last_name) == last_name.lower(),
-        func.lower(Attendee.email) == email.lower(),
+        Attendee.normalized_email == Attendee.normalize_email(email),
         Attendee.zip_code == zip_code
     ).first()
 
--- a/panels/site_sections/attractions.py
+++ b/panels/site_sections/attractions.py
@@ -19,6 +19,15 @@ def _attendee_for_badge_num(session, badge_num, options=None):
     return query.first()
 
 
+def _attendee_for_info(session, first_name, last_name, email, zip_code):
+    return session.query(Attendee).filter(
+        func.lower(Attendee.first_name) == first_name.lower(),
+        func.lower(Attendee.last_name) == last_name.lower(),
+        func.lower(Attendee.email) == email.lower(),
+        Attendee.zip_code == zip_code
+    ).first()
+
+
 def _model_for_id(session, model, id, options=None, filters=[]):
     if not id:
         return None
@@ -114,10 +123,19 @@ class Root:
             'badge_num': attendee.badge_num}
 
     @ajax
-    def signup_for_event(self, session, badge_num, id, **params):
-        attendee = _attendee_for_badge_num(session, badge_num)
-        if not attendee:
-            return {'error': 'Unrecognized badge number: {}'.format(badge_num)}
+    def signup_for_event(self, session, id, badge_num='', first_name='',
+                               last_name='', email='', zip_code='', **params):
+        if badge_num:
+            attendee = _attendee_for_badge_num(session, badge_num)
+            if not attendee:
+                return {
+                    'error': 'Unrecognized badge number: {}'.format(badge_num)
+                }
+        else:
+            attendee = _attendee_for_info(session, first_name, last_name,
+                                                   email, zip_code)
+            if not attendee:
+                return {'error': 'No attendee is registered with that info'}
 
         event = _model_for_id(session, AttractionEvent, id)
         if not event:
--- a/panels/templates/attractions/attractions_common.html
+++ b/panels/templates/attractions/attractions_common.html
@@ -20,6 +20,10 @@
     .visible-xs-inline { display: none !important; }
     .visible-xs-inline-block { display: none !important; }
     .hidden-xs { display: initial !important; }
+    .attendee-info-row input {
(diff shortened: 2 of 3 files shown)


EliAndrewC:
Good catch!  I totally forgot about ``Attendee.normalized_email``, but I've just made that change and re-tested locally to make sure it works.

RobRuana (approved):
I am into this! I have a few thoughts, but I approve this as-is if you don't have time to make any more changes.

1. The naming for "early signup" could be swapped for "alt signup", as someone at the event could still use this method (for example if they left their badge in their room)
2. If we wanted to enable this per-attraction, I'd suggest a boolean column on the attraction table, like `enable_early_signups`. And then in the UI we would hide the "I don't have my badge" button if we're pre-fest and `enable_early_signups` is `False`. We'd probably also want to add a model validation on the backend. (Enabling per-attraction is actually a non-trivial amout of work)
3. I added the ability to hide attraction signups from the public, so it's probably less important that we enable this on a per-attraction basis. Admins can keep their attractions hidden from the public until the event starts 😄
4. After the signup is successful, we should probably still show the done modal: `$doneModal.modal('show');`. Maybe pass a varible to it that hides the "sign up a friend" form?

That's all I've got. Again, if you're short on time, I'm happy to make additional changes after the merge!

RobRuana:
Also, from your screenshots, I'm glad to know the donut emoji "🍩" renders on linux!

panels/site_sections/attractions.py
```diff
@@ -19,6 +19,15 @@ def _attendee_for_badge_num(session, badge_num, options=None):
     return query.first()
 
 
+def _attendee_for_info(session, first_name, last_name, email, zip_code):
+    return session.query(Attendee).filter(
+        func.lower(Attendee.first_name) == first_name.lower(),
+        func.lower(Attendee.last_name) == last_name.lower(),
+        Attendee.normalized_email == Attendee.normalize_email(email),
+        Attendee.zip_code == zip_code
+    ).first()
```
kitsuta: We already have this -- it's `session.lookup_attendee` at https://github.com/magfest/ubersystem/blob/master/uber/models/__init__.py#L705. That version also handles duplicate attendees, which as we found can be pretty important. Looks like that version could stand to use normalized_email though.

panels/templates/attractions/events.html
```diff
@@ -134,13 +134,18 @@
...
+    });
+
     $signupModal.find('.confirm-row .btn-success').on('click', function (event) {
       event.preventDefault();
       var eventId = $signupModal.data('eventId'),
           $badgeNum = $signupModal.find('input[name=badge_num]'),
           badgeNum = $badgeNum.val();
 
-      signupForEvent(eventId, badgeNum, function(response) {
+      signupForEvent(eventId, {badge_num: badgeNum}, function(response) {
```
kitsuta, line 148: This is cool, but from a quick Google search it looks like it only works in the latest Firefox. That's a bit too bleeding-edge -- unless I'm chasing the wrong lead and this isn't a destructuring assignment.

kitsuta (changes requested):
This is super cool! I had a question about one of the JavaScript syntaxes you used, and we'd be better off if we used the existing attendee_lookup function, but I like this!

IMO, I think having attractions be private/public is good for now, and we don't need a config option for at-event/pre-event. In the future we can have a way for admins to schedule an attraction opening, which I think would completely cover the at-event-only use-case (and more!).

panels/templates/attractions/events.html
EliAndrewC: Ah, I see why you'd think that, but that's not what's happening here.

This is actually just passing an object literal as a parameter to this argument.  For example, instead of saying

```javascript
var d = {foo: 'bar'};
someFunc('baz', d, 'baf');
```

We can just say

```javascript
someFunc('baz', {foo: 'bar'}, 'baf'});
```

I can see why it looked like I was doing destructuring assignment, since what I did here looks a lot like the "Setting a function parameter's default value" section at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

However, in this case I'm only calling the function.  If I did something like this while *declaring* a function then you're right that it would only work in bleeding edge browsers and we should (sadly) avoid that kind of thing.

kitsuta: Ahh, yes that makes sense. It turns out googling "Javascript curly braces" isn't guaranteed to give context-aware results. Cool! Carry on!

panels/site_sections/attractions.py
EliAndrewC: Good catch, I've just made and tested this change.  You're right that version should use ``normalized_email``, so I've made that change at https://github.com/magfest/ubersystem/pull/3059

panels/site_sections/attractions.py
```diff
@@ -19,6 +19,13 @@ def _attendee_for_badge_num(session, badge_num, options=None):
     return query.first()
 
 
+def _attendee_for_info(session, first_name, last_name, email, zip_code):
+    try:
+        return session.lookup_attendee(first_name, last_name, email, zip_code)
+    except:
+        return None
```
EliAndrewC, line 26: As per @kitsuta 's suggestion I've updated this function to use ``session.lookup_attendee``.

panels/templates/attractions/attractions_common.html
```diff
@@ -208,6 +212,10 @@
     display: inline-block;
   }
 
+  .hidden-soldout.hidden-modal-form {
+    display: none;
+  }
```
EliAndrewC, line 217: I implemented @RobRuana 's suggestion about showing the ``$doneModal`` instead of just a toast and hiding the badge form if we got there from the "I don't know my badge number" form.  I implemented that by adding an ``addClass`` / ``removeClass`` which sets or removes the ``hidden-modal-form`` class appropriately.

panels/templates/attractions/events.html
```diff
@@ -75,6 +75,7 @@
   $(function() {
     var $modals = $('.modal'),
         $signupModal = $('#signup_modal'),
+        $altSignupModal = $('#alt_signup_modal'),
```
EliAndrewC, line 78: I liked @RobRuana 's suggestion of "alt" signup rather than "early" signup, so I renamed this to do that.

EliAndrewC:
I've made all of the suggested changes except for the optional extra config.  In the interest of time I'm gonna merge this in and deploy it tonight since the tabletop folks have been asking for it.  If we find a need for that extra config between now and the event, let me know and I'll jump on it ASAP.

PR #140, opened 2017-12-21
Adds attendee facing attraction signup management screen

RobRuana:
Depends on https://github.com/magfest/ubersystem/pull/3060

**NOTE**: The tab bar only shows up if you've actually signed up for any attractions.

# Prereg Confirmation With Tab Bar
screen shot 2017-12-21 at 3 57 05 am

# Attraction Signups Management
screen shot 2017-12-21 at 3 56 35 am

### DIFF
--- a/panels/models/attraction.py
+++ b/panels/models/attraction.py
@@ -4,6 +4,7 @@ import re
 from collections import OrderedDict
 from datetime import datetime, timedelta
 
+from sideboard.lib import listify
 from sideboard.lib.sa import JSON, CoerceUTF8 as UnicodeText, UTCDateTime, UUID
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import backref
@@ -33,6 +34,30 @@ def sluggify(s):
     return RE_SLUG.sub('-', s).lower().strip('-')
 
 
+def treeify(items, keys, value_key=None):
+    if not keys:
+        return items
+    keys = listify(keys)
+    last_key = keys[-1]
+    treeified = OrderedDict()
+    for item in items:
+        current = treeified
+        for key in keys:
+            attr = key(item) if callable(key) else getattr(item, key)
+            if attr not in current:
+                current[attr] = [] if key is last_key else OrderedDict()
+            current = current[attr]
+        if value_key:
+            if callable(value_key):
+                value = value_key(item)
+            else:
+                value = getattr(item, value_key)
+        else:
+            value = item
+        current.append(value)
+    return treeified
+
+
 @Session.model_mixin
 class Attendee:
     attraction_signups = relationship(
@@ -47,6 +72,15 @@ class Attendee:
     @property
     def attractions(self):
         return list({e.feature.attraction for e in self.attraction_events})
+
+    @property
+    def signups_by_attraction_by_feature(self):
+        signups = sorted(self.attraction_signups, key=lambda s: (
+            s.event.feature.attraction.name,
+            s.event.feature.name))
+        return treeify(signups, [
+            lambda s: s.event.feature.attraction,
+            lambda s: s.event.feature])
 
     def is_signed_up_for_attraction(self, attraction):
         return attraction in self.attractions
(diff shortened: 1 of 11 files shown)


panels/models/attraction.py
```diff
@@ -33,6 +34,30 @@ def sluggify(s):
...
+def treeify(items, keys, value_key=None):
+    if not keys:
+        return items
+    keys = listify(keys)
+    last_key = keys[-1]
+    treeified = OrderedDict()
+    for item in items:
+        current = treeified
+        for key in keys:
+            attr = key(item) if callable(key) else getattr(item, key)
```
kitsuta: Do you think it's a good idea to add a default to the `getattr` here, like `None`?

panels/models/attraction.py
```diff
@@ -33,6 +34,30 @@ def sluggify(s):
...
+            if attr not in current:
+                current[attr] = [] if key is last_key else OrderedDict()
+            current = current[attr]
+        if value_key:
+            if callable(value_key):
```
kitsuta: It's a little strange that you did this all on one line for `attr` and then did it differently here. For the record, I prefer the one line version.

panels/models/attraction.py
```diff
@@ -33,6 +34,30 @@ def sluggify(s):
     return RE_SLUG.sub('-', s).lower().strip('-')
 
 
+def treeify(items, keys, value_key=None):
```
kitsuta: Didn't you do something similar to this for searching the schedule? Or am I misremembering, somehow...?

panels/models/attraction.py
```diff
@@ -33,6 +34,30 @@ def sluggify(s):
...
+    if not keys:
+        return items
+    keys = listify(keys)
+    last_key = keys[-1]
+    treeified = OrderedDict()
+    for item in items:
+        current = treeified
```
kitsuta: Wait, are you.. transmuting the original OrderedDict() by assigning it to a second variable, which you then modify? If so, that really obfuscates what's happening. If not, I have no idea how `treeified` is ever modified.

panels/models/attraction.py
```diff
@@ -33,6 +34,30 @@ def sluggify(s):
...
+        for key in keys:
+            attr = key(item) if callable(key) else getattr(item, key)
+            if attr not in current:
+                current[attr] = [] if key is last_key else OrderedDict()
```
kitsuta: If I'm reading this all correctly, there's only one set of branching points, right? Everything else linearly digs down. So the data structure looks like:

```  ---
 / | \
/  |  \
   |   \
        \
```

is that correct?

kitsuta (approved):
Looks good! The treeify function is a little difficult to parse, but everything else looks solid and I didn't see anything that would be a big problem.

panels/models/attraction.py
RobRuana: For the schedule I made a `trie` which _is_ a tree, but a with a particular construction. This is _also_ making a tree, but with a different construction. Although I think a better name would be `groupify`, because that's the actual intent of the function. I'll make that change and add some docs!

😄

panels/models/attraction.py
RobRuana: I'm making _nested_ OrderedDicts. So when we start the iteration, the initial OrderedDict is the current one. As we go deeper in the tree, `current` will point to child OrderedDicts.

panels/models/attraction.py
RobRuana: I think we want this to raise an `AttributeError` if someone provides the function with an invalid key.

panels/models/attraction.py
RobRuana: I'll add some docs, but this function can nest arbitrarily deep.

panels/models/attraction.py
RobRuana: Yeah, that was because of my self-imposed 79 character limit. I'll rework it to be on one line!

panels/models/attraction.py
```diff
@@ -33,6 +34,159 @@ def sluggify(s):
     return RE_SLUG.sub('-', s).lower().strip('-')
 
 
+def groupify(items, keys, val_key=None):
+    """
+    Groups a list of items into nested OrderedDicts based on the given keys.
```
kitsuta, line 39: Whoa! With this kind of flexibility, this could be its own library! Thank you for the docs!

PR #141, opened 2017-12-22
Improves schedule render performance by hiding the main container and preventing reflows

RobRuana:


thaeli:
LGTM

panels/templates/schedule/edit.html
```diff
@@ -305,6 +305,8 @@
     };
 
     $(function(){
+        $('#mainContainer').css('display', 'none');
```
kitsuta, line 308: Highly recommend a comment gets added at some point soon explaining this change -- I can see someone in the future going "Why is this here? Better remove it"

PR #142, opened 2017-12-26
Adds support for attraction notifications and attendee facing signup management

RobRuana:
screen shot 2017-12-26 at 8 36 26 am


RobRuana:
Requires https://github.com/magfest/ubersystem/pull/3064

panels/config.py
```diff
@@ -86,15 +86,15 @@ def _make_room_trie(rooms):
...
-    schedule_menu = [MenuItem(name='View Schedule', href='../schedule/internal')]
+    schedule_menu = [MenuItem(name='View Schedule', access=c.STUFF, href='../schedule/internal')]
 
 schedule_menu.extend([
-    MenuItem(name='Edit Schedule', href='../schedule/edit'),
+    MenuItem(name='Edit Schedule', access=c.STUFF, href='../schedule/edit'),
     MenuItem(name='Attractions', href='../attractions_admin/')])
 
-c.MENU.submenu.insert(2, MenuItem(name='Schedule', access=c.STUFF, submenu=schedule_menu))
+c.MENU.submenu.insert(2, MenuItem(name='Schedule', access=[c.STUFF, c.PEOPLE, c.REG_AT_CON], submenu=schedule_menu))
```
RobRuana, line 97: More fine grained access restrictions on Schedule menu, to allow reg people to check attendees into attractions.

panels/models/attraction.py
```diff
@@ -1,37 +1,36 @@
...
 import re
 
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 from datetime import datetime, timedelta
 
 from sideboard.lib import listify
 from sideboard.lib.sa import JSON, CoerceUTF8 as UnicodeText, UTCDateTime, UUID
+from sqlalchemy import and_, exists, func, or_, select, text, union, not_, cast, alias
+from sqlalchemy.ext.associationproxy import association_proxy
```
RobRuana, line 11: Using an AssociationProxy is one way we might be able to fix [this issue](https://github.com/magfest/ubersystem/issues/3020). See the SQLAlchemy docs on [Simplifying Association Objects](http://docs.sqlalchemy.org/en/latest/orm/extensions/associationproxy.html#simplifying-association-objects). 

By using an AssociationProxy, I believe it forces the association object to be loaded into the session, thus recorded in the tracking table.

panels/models/attraction.py
```diff
@@ -444,6 +562,8 @@ def available_events_by_day(self):
 # =====================================================================
 class AttractionEvent(MagModel):
     attraction_feature_id = Column(UUID, ForeignKey('attraction_feature.id'))
+    attraction_id = Column(UUID, ForeignKey('attraction.id'), index=True)
```
RobRuana, line 605: This is a design pattern I am testing out. The `Attractions` hierarchy looks like this:
```
Attraction
  + AttractionFeature
      + AttractionEvent
          + AttractionSignup
          + AttractionNotification
          + AttractionNotificationReply
```

I found I was doing a lot of expensive joins and traversals to get from a child object back to the `Attraction` object. I denormalized the `attraction_id`, so I could reach/join from any of the child objects directly to the parent `Attraction`.

This does have performance benefits, but it also requires extra attention to make sure the `attraction_id` is set correctly. This is mostly taken care of by using `AssociationProxys` and `@presave_adjustments`. But it's something to be aware of.

panels/models/attraction.py
```diff
@@ -569,28 +706,106 @@ def overlap(self, event):
...
 class AttractionSignup(MagModel):
     attraction_event_id = Column(UUID, ForeignKey('attraction_event.id'))
+    attraction_id = Column(UUID, ForeignKey('attraction.id'))
     attendee_id = Column(UUID, ForeignKey('attendee.id'))
 
-    signup_time = Column(UTCDateTime, server_default=utcnow())
-    checkin_time = Column(UTCDateTime, nullable=True)
+    signup_time = Column(UTCDateTime, default=lambda: datetime.now(pytz.UTC))
+    checkin_time = Column(
+        UTCDateTime, default=lambda: utcmin.datetime, index=True)
```
RobRuana, line 769: Here is another interesting design choice made for the sake of performance. I made `checkin_time` not null, so it could be indexed. Indexes on nullable columns don't work when you are trying to select rows where the column is null.

Instead of using null to indicate "not checked in yet", I am using the date of "0001-01-01 0:00". This has really sped up some queries, but we can no longer use the Falseyness of `checkin_time` to indicate "not checked in".

RobRuana:
@EliAndrewC or @kitsuta could either of you take a look at this sometime today? It's the last piece of the attractions feature, and it's also probably the most janky. And we're running out of time.

panels/automated_emails.py
```diff
@@ -1,7 +1,12 @@
...
 
-AutomatedEmail.queries[PanelApplication] = lambda session: session.query(PanelApplication) \
-    .options(subqueryload(PanelApplication.applicants).subqueryload(PanelApplicant.attendee))
+
+AutomatedEmail.queries[PanelApplication] = lambda session: session \
+    .query(PanelApplication) \
+    .options(
+        subqueryload(PanelApplication.applicants)
+            .subqueryload(PanelApplicant.attendee)) \
+    .order_by(PanelApplication.id)
```
EliAndrewC, line 9: What's the purpose of this order by?  Just to make the query deterministic on general principle?

panels/config.py
```diff
@@ -86,15 +86,15 @@ def _make_room_trie(rooms):
         schedule_name = 'View Schedule Externally'
     schedule_menu = [
         MenuItem(name=schedule_name, href='../schedule/'),
-        MenuItem(name='View Schedule Internally', href='../schedule/internal')]
+        MenuItem(name='View Schedule Internally', access=c.STUFF, href='../schedule/internal')]
```
EliAndrewC, line 89: Whoops, good catch.

panels/configspec.ini
```diff
@@ -19,6 +19,10 @@ music_rooms = string_list(default=list("concerts", "chiptunes", "pose_lounge", "
 # deprecated "peglegs_email_signature" from the main repo.
 panels_email_signature = string(default=" - MAGFest Panels Team")
 
+# Attractions Notifications
+panels_twilio_number = string(default="")
```
EliAndrewC, line 23: Normally I'd suggest that we make this more generic like ``panels_source_number`` in case we want to switch from Twilio, but I never ever want to switch from Twilio, since it's the best SMS provider EVER.  (I say that having now worked with about half a dozen of them for my day job.)

EliAndrewC:
I'll take a look this morning and should be able to give a +1 before lunch.

RobRuana:
A +1 or change requests. Either way! This code is pretty janky, and there's a distinct lack of unit tests.

panels/models/attraction.py
EliAndrewC: Nice!  I hadn't read up on AssociationProxy objects until now, and that's a great tool I'll keep in mind in the future.  I can already think of a few places in some code I've written at work that would benefit from that.

panels/models/attraction.py
```diff
@@ -187,13 +187,51 @@ def groupify(items, keys, val_key=None):
     return groupified
 
 
+def mask(s, mask_char='*', min_unmask=1, max_unmask=2):
```
EliAndrewC, line 190: A short docstring would be useful here.  It took me a minute to realize that this was basically used for replacing part of a string with asterisks for display, like for phone numbers and such.

panels/models/attraction.py
```diff
@@ -202,6 +240,26 @@ def attraction_features(self):
     def attractions(self):
         return list({e.feature.attraction for e in self.attraction_events})
 
+    @property
+    def masked_email(self):
+        name, _, domain = self.email.partition('@')
```
EliAndrewC, line 283: Nice, I always forget that ``str.partition`` exists.

panels/models/attraction.py
```diff
@@ -358,6 +426,57 @@ def locations(self):
     def locations_by_feature_id(self):
         return groupify(self.features, 'id', lambda f: f.locations)
 
+    def signups_requiring_notification(
+            self, session, from_time, to_time, options=None):
```
EliAndrewC, line 468: I recommend naming the ``options`` parameter something like ``queryset_options`` since I had to scroll down to see what this was actually used for.

panels/models/attraction.py
```diff
@@ -358,6 +426,57 @@ def locations(self):
...
+                        == AttractionSignup.attraction_event_id,
+                    AttractionNotification.attendee_id
+                        == AttractionSignup.attendee_id)))).with_labels()
+            subqueries.append(subquery)
+
+        query = subqueries[0].union(*subqueries[1:])
+        if options:
+            query = query.options(*listify(options))
+        query.order_by(AttractionSignup.id)
+        return groupify(query, lambda x: x[0], lambda x: x[1])
```
EliAndrewC, line 516: Nice - I had no idea what ``groupify`` was doing, but I scrolled up and found where it's declared and it has a really kickass docstring.  A+ would group again.

panels/models/attraction.py
```diff
@@ -488,24 +633,37 @@ def start_time_label(self):
...
+        advance_checkin = self.attraction.advance_checkin
+        if advance_checkin < 0:
+            return self.start_time
         else:
-            return self.start_time - timedelta(seconds=required_checkin)
+            return self.start_time - timedelta(seconds=advance_checkin)
+
+    @property
+    def checkin_time_local(self):
+        return self.checkin_time.astimezone(c.EVENT_TIMEZONE)
```
EliAndrewC: Do we no longer have a suffix property which automatically does this for all UTCDateTime column fields on ``MagModel``?

panels/models/attraction.py
EliAndrewC: Interesting.  I like this general idea - I'll have to think about that for some of my work projects.  We definitely have a lot of nullable timestamps, though I can't recall offhand whether any of them are indexed.

panels/models/attraction.py
```diff
@@ -187,13 +187,51 @@ def groupify(items, keys, val_key=None):
...
+
 @Session.model_mixin
 class Attendee:
+    NOTIFICATION_EMAIL = 0
+    NOTIFICATION_TEXT = 1
+    NOTIFICATION_NONE = 2
+    NOTIFICATION_PREF_OPTS = [
+        (NOTIFICATION_EMAIL, 'Email'),
+        (NOTIFICATION_TEXT, 'Text'),
+        (NOTIFICATION_NONE, 'None')]
```
EliAndrewC, line 254: What's the rationale for defining this enum here and in this way instead of adding an ``[enums]`` subsection like we usually do?

panels/notifications.py
```diff
@@ -0,0 +1,210 @@
...
+                    now='is right now',
+                    past_prefix='was ',
+                    past_suffix=' ago')
+
+            ident = AttractionEvent.get_ident(event.id, advance_notice)
+            if use_text:
+                type_ = Attendee.NOTIFICATION_TEXT
+                body = ATTRACTION_TPL.format(signup=signup, checkin=checkin)
+                subject = ''
+                sid = send_sms(attendee.cellphone, body)
```
EliAndrewC: Since ``send_sms`` can raise exceptions, do we want this entire function to stop if that happens?  I'm just worried that a single weird edge case will stop all of the other notifications from being sent.

I realize that ``send_sms`` usually returns ``'Unable to send sms'`` when it fails rather than raising an exception, but in those cases where it doesn't, then it stops the remaining notifications from being sent, which seems dangerous.

panels/notifications.py
```diff
@@ -0,0 +1,210 @@
...
+                    template = 'emails/attractions_notification.html'
+                    subject = 'Checkin for {} is at {}'.format(
+                        event.name, event.checkin_time_label)
+
+                body = render(template, {
+                    'signup': signup,
+                    'checkin': checkin,
+                    'c': c}).decode('utf-8')
+                sid = ident
+                send_email(
```
EliAndrewC: Same comment above w.r.t. ``send_email`` raising an exception sometimes.

panels/sep_commands.py
```diff
@@ -0,0 +1,172 @@
...
+# vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
+
+from sqlalchemy.engine.default import DefaultDialect
+from sqlalchemy.sql.sqltypes import String, DateTime, NullType
+
+# python2/3 compatible.
+PY3 = str is not bytes
+text = str if PY3 else unicode
+int_type = int if PY3 else (int, long)
+str_type = str if PY3 else (str, unicode)
```
EliAndrewC, line 16: I'm pretty sure this codebase is pretty solidly Python 3 only, so is this really necessary?

Even if it is necessary, is there any reason not to use ``six``?

panels/sep_commands.py
```diff
@@ -0,0 +1,172 @@
...
+    ).string
+    return RE_INTERVAL.sub(r"interval '\1 seconds'", s)
+
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+# DEVELOPMENT TOOLS - DEVELOPMENT TOOLS - DEVELOPMENT TOOLS - DEVELOPMENT TOOLS
+# =============================================================================
+
+
+if c.DEV_BOX:
+    @entry_point
```
EliAndrewC: Huh.  I'm not sure why it's never occurred to me to wrap an ``@entry_point`` inside this kind of if statement.  I've always done it in reverse, but this seems better.

panels/site_sections/attractions.py
```diff
@@ -197,10 +209,24 @@ def cancel_signup(self, session, attendee_id, id):
...
                 session.commit()
         if message:
             return {'error': message}
+
+    @ajax
+    def opt_out(self, session, id, attractions_opt_out):
+        if cherrypy.request.method == 'POST':
+            attendee = session.query(Attendee).get(id)
+            attendee.attractions_opt_out = attractions_opt_out
+            session.commit()
```
EliAndrewC, line 226: Although it wouldn't make any real difference here, I usually like returning something from Ajax methods, even if it's only an empty dict.  Not a big deal either way, but it's nice to be able to rely on an object being returned so that the client side can check for the presence of fields, etc.

panels/site_sections/attractions_admin.py
```diff
@@ -5,10 +5,11 @@
 
 @all_renderable(c.STUFF)
 class Root:
+    @renderable_override(c.STUFF, c.PEOPLE, c.REG_AT_CON)
     def index(self, session, filtered=False, message='', **params):
+        admin_account = session.current_admin_account()
         if filtered:
-            admin_account_id = cherrypy.session['account_id']
-            attraction_filter = [Attraction.owner_id == admin_account_id]
+            attraction_filter = [Attraction.owner_id == admin_account.id]
```
EliAndrewC, line 12: I can't figure out why Github is showing the above two lines as a change.  It looks like the exact same line with no changes is showing up as removed and re-added.  Am I missing something?

panels/templates/attractions/events.html
```diff
@@ -267,52 +266,49 @@ <h4 class="modal-title">
           <span class="bling-icon"></span> <span class="signup-event-label"></span>
         </h4>
       </div>
-      <div class="modal-body">
```
EliAndrewC, line 270: I look forward to seeing how this is now styled.  I figured you could do a much better job than me with making this modal look good - I spent an embarrassing amount of time getting it to look as decent as it did, but I was never really happy with it.

EliAndrewC (approved):
Looks great!

I had a few minor suggestions and questions, which can either be addressed now or after this is merged.

My only remaining question: is "Jesse Babylon" a reference to something?  I didn't recognize it and a quick Google search didn't turn up anything obvious.

panels/automated_emails.py
RobRuana: In short, `order_by` is required whenever you are using `subqueryload()`. Without deterministic ordering, you can end up with the wrong objects in your relationship attributes. From a performance perspective, it's negligible because `id` is an indexed column. I refer you the following docs:
* [The Importance of Ordering](http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#the-importance-of-ordering)
* [Why is ORDER BY required with LIMIT (especially with subqueryload())?](http://docs.sqlalchemy.org/en/latest/faq/ormconfiguration.html#why-is-order-by-required-with-limit-especially-with-subqueryload)

panels/configspec.ini
RobRuana: I think the entire notifications system is going to be reworked in 2018, so I'm not devoting a huge amount of time to design or planning for this particular pull request.

panels/models/attraction.py
RobRuana: It's self documenting!

panels/models/attraction.py
RobRuana: `groupify()` will eventually be moved to a more appropriate place. Like in `pockets` maybe?

panels/models/attraction.py
RobRuana: Hah! I didn't even know that existed! I will remove these properties. I have a real love/hate relationship with `@suffix_property`. On the one hand, it's super convenient. On the other hand, our implementation means that any errors in a property are completely swallowed and hidden. See the following issues for more discussion:
* https://github.com/magfest/ubersystem/issues/1942
* https://github.com/magfest/ubersystem/issues/2842

panels/notifications.py
RobRuana: Hmmm... yes, I'll wrap these in exception handlers.

panels/sep_commands.py
RobRuana: This is some bullshit I copy/pasted. This entire file needs to be removed. The whole thing is a hack job.

panels/site_sections/attractions_admin.py
RobRuana: `admin_account_id` vs `admin_account.id`

Note the underscore (`_id`) vs the period (`.id`).

panels/templates/attractions/events.html
RobRuana: Visually, it looks _almost_ exactly the same. But it does the right thing on mobile screens (uses the full width vs a fixed width/fixed margin).

One big change I made was to move the `modal-footer` inside the `form` element. Additionally, instead of hooking the `button.click` event, I am hooking the `form.submit` event. I call `event.preventDefault()`, so we don't actually POST the form – we still use AJAX to post the data the way we did before.

Moving the `model-footer` inside the `form` element and hooking `form.submit` instead of `button.click` has several cool benefits:
* It allows you to hit the enter key in a text field to submit the data instead of requiring a mouse click on the button, or tabbing the button into focus
* It gets HTML5 form validation by the browser for free. The browser intercepts the `form.submit` event before it gets to our handler, and does client-side form validation (like "Fill out this field", see screenshot below).

Also, I removed the rounded corners from the middle fields.

# Old: Rounded Corners on Internal Fields
screen shot 2017-12-28 at 6 53 35 am

# New: Sharp Corners on Internal Fields
screen shot 2017-12-28 at 7 04 09 am

RobRuana:
You have yourself to thank for Jesse Babylon! It's a fake name generated for your test data! See line 247461 of [test_data.json](https://github.com/magfest/ubersystem/blob/master/uber/tests/test_data.json).

panels/models/attraction.py
RobRuana: In short, it's because I'm lazy and most of this will be redone next year.

Going into more detail, I have a real love/hate relationship with `[enums]` (much like I do with `@suffix_property`). I have a much larger rant about `[enums]`, but instead of wailing and gnashing my teeth, I want to talk a little bit about the distinction between code, constants, and configuration.

Presumably, you'd like `NOTIFICATION_PREFS` to exist in the _config_ because you'd like it to be _configurable_, right? Not every deployment will have a twilio account (or will otherwise be able to send text messages). So you'd like to _configure_ that on a per deployment basis, right?

Here's the thing though. In this implementation, I explicitly reference the individual enum values: `NOTIFICATION_EMAIL`, `NOTIFICATION_TEXT`, and `NOTIFICATION_NONE`. So, lets pretend we've implemented this in an `[enums]` section of the config. What happens when we add a new value to the enum?

Let's pretend we add `NOTIFICATION_PIGEON`, because we want to be able to notify attendees using carrier pigeon. What happens? Well, nothing happens. The code is not aware of the existence of `NOTIFICATION_PIGEON`. It's behavior will not change based on the addition of that value.

Instead of _adding_ a value to the enum, let's consider what happens when we _remove_ a value from the enum. Consider a deployment that does not have a twilio account, and cannot send text messages. We'd like to entirely disable text messages. What happens when we remove `NOTIFICATION_TEXT` from the enum?

Everything breaks, of course. References to `NOTIFICATION_TEXT` are hardcoded in the source code. As soon as those lines of code are executed, the program will fail catastrophically. Due to the way this feature has been written, we _cannot_ remove `NOTIFICATION_TEXT`. The code relies the existence of that value.

To put it another way, it is not _configurable_. It is not a thing we can configure. It does not belong in the configuration. It is a constant that belongs in the code.

Now, I know what you're thinking. "But we _do_ want to configure this! This is precisely the kind of thing that we _want_ to be able to configure!" And I agree! However, we should not delude ourselves into believing that storing these values in the configuration immediately makes them configurable. It does not! If we put these values into the config – without changing our implementation at all – then all we've accomplished is to mix our source code with our configuration.

You may as well be asking why the various [socket options](https://github.com/freebsd/freebsd/blob/master/sys/sys/socket.h#L102) are declared in source code instead of in a configuration file.

It's because the socket implementation has an inherent understanding of each of those constants. Those constants have an explicit meaning to the source code. The source code expects those constants to exists, and it would break if they didn't.

A good rule of thumb to follow when deciding whether or not to use `[enums]` is to ask yourself if the code will break when a value is removed. This implementation will certainly break, because it expects each of those values to exist, so I kept it out of the configuration.

# There has to be a Better Way<sup>TM</sup>

But back to the original point, we definitely _want_ the available notification options to be configurable. So what's the solution?

This is where we start talking about dependency injection. What we really want to do is create a well-defined interface, like `NotificationProvider` or something. And then in our config, we can add or remove `NotificationProviders` which _will_ change the behavior of the source code. We don't need to know about the internals of a `NotificationProvider`, only what the interface defines.

A good example of this that we're all familiar with is the python logging system. You configure log handlers by adding them in your config file. You don't need to understand the internals of a log handler. Some of them write to a file. Some of them send data over the network. You might even have a log handler that emails you. Someone in the future might write a log handler that we couldn't possibly imagine, and it will still work without modifying the current codebase.

So what you're really talking about is being able to configure our notification preferences the way we configure our logging. And that's what I'd love to do if we had more time. But to quote [Devops Rumsfeld](https://twitter.com/agentdero/status/200372798424760320):
> You go to production with the code you have, not the code you wish you had

I'm going to take it a step further, though. I'm going to say that even in a perfect world where we had the time to do this feature right, we still wouldn't want to put it in the config file. If you step back and think about it, we are not configuring the _system_ as a whole. We are configuring a single _event's_ ability to send notifications. Perhaps Super MAGFest has the budget to support text messaging, but MAGStock does not. Or maybe text messaging just doesn't make sense for MAGStock because we're in the woods and there's no cellular reception (_I actually don't think that's true_).

In any case, we'd want to be able to configure notification options _per event_ not _per system_. In which case, we'd want to store notification options in the _database_, not the configuration file. We might still keep a list of _available_ `NotificationProviders` in the config. But even then, I'd argue we'd be better off importing them dynamically, perhaps from plugins, and using introspection to discover what `NotificationProviders` are available.

panels/models/attraction.py
RobRuana: Indexing nullable columns still works if you're querying for a non-null value. For example, if you want to query for rows matching a given window, between start and end dates, the database will be able to use the index. However, if you're querying for rows with a null value, the database will have to do a table scan.

I am actually quite surprised by this behavior, and I only discovered it by running some queries against `EXPLAIN` in `psql`. I would have assumed that postgres would have some way of knowing which rows where _not_ contained in the index. Like, somehow the index would keep all the rows with null values in a set that it could return if you're querying for a null value.

panels/site_sections/attractions_admin.py
EliAndrewC: Ahhh, I figured it was something like that but I stared at that line for like 30 seconds and still didn't see it.  Derp.

panels/models/attraction.py
EliAndrewC: I definitely also assumed that null values on an indexed column would still be queryable without a full table scan - glad to know that's not the case!

panels/automated_emails.py
EliAndrewC: Ahh, good to know.  I usually end up using ``joinedload``, but I'll definitely have to keep this in mind for those occasions where I have enough levels of nested relationships that I need subqueryload.

RobRuana: I know each `subqueryload` adds an additional `select` statement, but the queries generated by `subqueryload` are almost always more efficient than the one's generated by `joinedload`. I know there's a way to speed up `joinedloads` by using `innerjoin=True`, if you're loading a non-nullable relationship (as [described here](http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#joined-eager-loading)). I've never experimented with that, but maybe I should give it a try!

EliAndrewC: Hmm, interesting.  I've recently implemented some automated performance measurements on a recent microservice we implemented at work, so that would be a good opportunity to try switching from ``joinedload`` to ``subqueryload`` and see what effect that has on the overall performance of the API, since the most performance-critical pieces all have some kind of ``joinedload``.

Thanks for the tip!

RobRuana: Definitely let me know the results (and if `innerjoin=True` makes a difference)!

panels/models/attraction.py
EliAndrewC: Fair enough!

FWIW, ``[enums]`` came about for somewhat different reasons that you seem to think.  To be honest, if I'd known as much about metaclasses then as I do now then I would have done things a lot differently.

What I noticed was that I had a lot of code in the old days that looked like this:

```python
SMALL = 1
MEDIUM = 2
LARGE = 3
SIZE_OPTS = (
    (SMALL, 'small'),
    (MEDIUM, 'medium'),
    (LARGE, 'large')
)
SIZES = dict(SIZE_OPTS)
```

So what's wrong with that?  It's not fundamentally bad code or anything, but it's twice as long as it needs to be.  I almost never care about the actual integer values of my enum options, but I still end up having to devote two lines of code instead of one to declaring them.  I also wanted to be able to automatically have both an ordered list/tuple version and a dictionary option without having to add a line of code every single time.

In other words, I wanted a declarative way to generate this kind of config.  If I had to do it over again, I would have used a metaclass to do something like

```python
class Sizes(ConfigEnum):
    SMALL = 'small'
    MEDIUM = 'medium'
    LARGE = 'large'
```

With the right metaclass, I could have done almost anything I wanted, including:
- Automatically generating integer constants for each field.
- Automatically generating fields to meet our current use cases of having things like ``c.SIZE_OPTS``, ``c.SIZES``, and ``c.SIZE_VARS``
- Overriding the defined-in-code values in a config file if I wanted.  For example, suppose we wanted to change ``'medium'`` to ``'middle'``.  We could have a special ``[enums]`` subsection called ``[[sizes]]`` in our config file with the ability to override the default text, which would still give us the "just edit a config file instead of having to edit the code to change a label" feature.

Even before I knew about metaclasses, the main thing which stopped me from trying to hack something together using this approach was ordering.  We usually want our ``_OPTS`` to be in the order in which they were declared.  There are a couple of ways we could have done that, e.g.

1) Use introspection of the source code to figure out what line the value was declared on.  That's pretty janky and magicky (even for me!), so I didn't want to do anything like that.

2) Instead of using raw strings, wrap your strings in some kind of function call which can increment a counter behind the scenes, e.g.

```python
class Sizes(ConfigEnum):
    small = val('small')
    medium = val('medium')
    large = val('large')
```

but I really hated how that looked.

Thankfully now we have the ``enum`` module, which preserves ordering, so if ``Sizes`` was an enum subclass then we would know the order in those options were declared.  (Enums already have a metaclass they use for a lot of their functionality, and I'm not sure how easy it is to hook into that, but I think it would still give me a way to solve the most annoying part of my use case.)

Anyway, the TLDR of all of this is that ``[enums]`` was really less about "everything should just be configurable" and more about "Eli hates code duplication and wants a declarative way to define these kinds of options".

There are obviously some intended benefits like being able to change the textual label for an enum field.  I also liked being able to define a base set of things like badge types and then add new options onto that in config, while allowing the configspec to mandate that the "everything breaks if you don't have this" options are still there.

I get what you're saying though.  While I still like my overall approach to ``[enums]`` from a "good declarative way to get what we need" perspective, it creates a tight coupling of code and config which is pretty bad from a maintainability perspective.  It's one of a number of design decisions that probably stopped being correct as soon as we started actually using these config files as, like, actual config files.

(Though as a philosophical issue, I generally think it's better to think of a ``configspec.ini`` as being *code* whereas a ``development-defaults.ini`` is obviously *config*.  I realize not everyone agrees with this, since a configspec is by definition a config file, but I think it's more accurate to think of it as code.  Not that this matters much to your overall point.)

Regardless, you'll certainly have my blessing to tear up the ``[enums]`` code whenever you get to a point where it makes sense to do that.  I'll be disappointed if we end up with a two-lines-of-code-instead-of-one but that's not a big deal in the grand scheme of things.

RobRuana: If we're just trying to cut down on boilerplate, I think we could solve that problem with some pretty straightforward code (ignoring the fact that [python now has enum](https://docs.python.org/3/library/enum.html)). No magic involved:

```python
>>> class enum:
...     def __init__(self, *args, **kwargs):
...         self._VARS = args
...         self._OPTS = []
...         count = 0
...         for arg in args:
...             setattr(self, arg, count)
...             self._OPTS.append((count, kwargs.get(arg, arg.title())))
...             count += 1
...         self._LABELS = dict(self._OPTS)
... 
>>> class Attendee:
...     NOTIFICATION_PREFS = enum('NONE', 'EMAIL', 'TEXT', TEXT='Text/SMS')
... 
>>> Attendee.NOTIFICATION_PREFS._VARS
('NONE', 'EMAIL', 'TEXT')
>>> 
>>> Attendee.NOTIFICATION_PREFS._OPTS
[(0, 'None'), (1, 'Email'), (2, 'Text/SMS')]
>>> 
>>> Attendee.NOTIFICATION_PREFS._LABELS
{0: 'None', 1: 'Email', 2: 'Text/SMS'}
>>> 
>>> Attendee.NOTIFICATION_PREFS.NONE
0
>>> Attendee.NOTIFICATION_PREFS.EMAIL
1
>>> Attendee.NOTIFICATION_PREFS.TEXT
2
```

EliAndrewC: I definitely considered something like that, and I did in fact have an ``enum`` class that was similar to that many years ago.  However, in the (fairly common) case where you want a custom label for all of your options, your implentation still requires us to write each option twice, once for ordering and once for labeling.  (Thankfully, Python has ``enum`` now which handles the ordering for us.)

In general I think we're in agreement though, since I think we both have roughly the same opinion that ``[enums]`` is worse than this general approach of having a declarative class.

RobRuana: OMG, you're killin' me brah!

```python
>>> import six
>>> class enum:
...     def __init__(self, *args):
...         self._VARS = []
...         self._OPTS = []
...         count = 0
...         for arg in args:
...             if isinstance(arg, six.string_types):
...                 label = arg.replace('_', ' ').title()
...             else:
...                 label = arg[1]
...                 arg = arg[0]
...             setattr(self, arg, count)
...             self._VARS.append(arg)
...             self._OPTS.append((count, label))
...             count += 1
...         self._LABELS = dict(self._OPTS)
... 
>>> NOTIFICATION_PREFS = enum('NONE', 'EMAIL', 'CARRIER_PIGEON', ('TEXT', 'Text/SMS'))
>>> 
>>> NOTIFICATION_PREFS._VARS
['NONE', 'EMAIL', 'CARRIER_PIGEON', 'TEXT']
>>> 
>>> NOTIFICATION_PREFS._OPTS
[(0, 'None'), (1, 'Email'), (2, 'Carrier Pigeon'), (3, 'Text/SMS')]
>>> 
>>> NOTIFICATION_PREFS._LABELS
{0: 'None', 1: 'Email', 2: 'Carrier Pigeon', 3: 'Text/SMS'}
>>> 
>>> NOTIFICATION_PREFS.NONE
0
>>> NOTIFICATION_PREFS.EMAIL
1
>>> NOTIFICATION_PREFS.CARRIER_PIGEON
2
>>> NOTIFICATION_PREFS.TEXT
3
```

EliAndrewC: Haha, I was honestly thinking of trolling you by saying "You forgot to do a ``.replace('_', ' ')`` for the description :P" and then I saw that you literally did make that improvement between your last version and this one, along with Python 2/3 compatibility for some reason :rofl: 

I do really like this approach though.  I would totally support doing something like this.  (I'd probably replace the autoincrementing ``count`` with the hash-based approach we're using for ``[enums]`` though, since otherwise you wouldn't be able to insert new items in the middle without changing the values of all of the existing items which come after that.)

I will say that I was sad that one of the earlier ``enum`` proposals got dropped, which would have allowed us to specify enum options without assignment, e.g.

```python
class NotificationPrefs(ConfigEnum):
    NONE
    CARRIER_PIGEON
    TEXT = 'Text/SMS'
```

but they eventually decided that violated Python's distinction between expressions and statements.  Which is fair, but that's pretty much the only thing I could ask for that would make this better than your overall approach to ``enum`` :)

PR #143, opened 2017-12-28
Move barcode scanning detection to barcode plugin

kitsuta:
Requires https://github.com/rams/barcode/pull/21.

### DIFF
--- a/panels/templates/attractions_admin/checkin.html
+++ b/panels/templates/attractions_admin/checkin.html
@@ -95,8 +95,12 @@
 
       {% set detect_barcode_scan = True %}
 
+      var setBarcodeField = function () {
+          $barcodeField = $('input[name=badge_num]');
+      };
+
       var useBarcode = function (badgeNum) {
-          getSignups(badgeNum)
+          getSignups(badgeNum);
       };
 
     var updateAttendee = function(attendee) {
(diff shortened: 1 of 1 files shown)


panels/templates/attractions_admin/checkin.html
```diff
@@ -93,36 +93,11 @@
...
-            } else if(response['badge_num']) {
-              var badgeNum = response['badge_num'];
-              $badgeNum.val(badgeNum);
-              getSignups(badgeNum);
-            }
-          }
-        });
-      }
-    });
+      {% set detect_barcode_scan = True %}
```
RobRuana, line 96: Does this actually work? Wouldn't all of the code in `baseextra.html` already be rendered by the time we reach this line (meaning `detect_barcode_scan` would be `False` at the time the barcode plugin `{% extends "baseextra.html" %}`)?

I'm actually not sure! I'm going to pull this branch locally and test it right now!

kitsuta: Oh shoot, that's a very good point. I'll try to brainstorm another tactic. Let me know the testing results!

kitsuta: Did you ever get this tested?

RobRuana: Oh, no! Fell off my radar ☹️

PR #148, opened 2017-12-29
Needed to also remove delete cascade

RobRuana:


kitsuta (commented):
I just saw the last PR and was really confused, so I'm glad I remembered correctly. 😅 +1 (belatedly)

PR #157, opened 2018-01-03
Prevents alternate attraction signups during AT_THE_CON.

RobRuana:
Also, always sends the welcome email for initial attractions signups.

### DIFF
--- a/panels/site_sections/attractions.py
+++ b/panels/site_sections/attractions.py
@@ -151,7 +151,9 @@ class Root:
     @ajax
     def signup_for_event(self, session, id, badge_num='', first_name='',
                          last_name='', email='', zip_code='', **params):
-        if badge_num:
+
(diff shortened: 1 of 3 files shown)


panels/site_sections/attractions.py
```diff
@@ -151,7 +151,9 @@ def verify_badge_num(self, session, badge_num, **params):
     @ajax
     def signup_for_event(self, session, id, badge_num='', first_name='',
                          last_name='', email='', zip_code='', **params):
-        if badge_num:
+
+        # Badge number during AT_THE_CON is a hard requirement for Autographs
```
EliAndrewC, line 155: Ahhh, good idea.  I originally figured there was no harm in allowing badgeless signups even during the event, but that would potentially be really bad for autographs.

EliAndrewC (approved):

PR #158, opened 2018-01-05
Make "Anytime during event" actually do what it says

kitsuta:
We have a config option for attractions for when check-in is required. Even though we had two separate options for "Anytime during event" and "When the event starts," they were actually doing the same thing. This makes them do what they say on the tin.

### DIFF
--- a/panels/models/attraction.py
+++ b/panels/models/attraction.py
@@ -666,7 +666,7 @@ class AttractionEvent(MagModel):
     def checkin_time(self):
         advance_checkin = self.attraction.advance_checkin
         if advance_checkin < 0:
-            return self.start_time
+            return self.end_time
         else:
             return self.start_time - timedelta(seconds=advance_checkin)
 


panels/models/attraction.py
```diff
@@ -666,7 +666,7 @@ def start_time_label(self):
     def checkin_time(self):
         advance_checkin = self.attraction.advance_checkin
         if advance_checkin < 0:
-            return self.start_time
+            return self.end_time
         else:
             return self.start_time - timedelta(seconds=advance_checkin)
```
kitsuta, line 671: We don't have to write up a separate `elif` clause for option 0 ("When the event starts") because start_time minus 0 = start_time.

EliAndrewC (approved):

kitsuta:
PS: Autographs wanted this, and they're the only Attraction on prod that should be affected.

PR #159, opened 2018-01-05
Emergency attractions fix, differentiates an event's checkin_start_time from checkin_end_time

RobRuana:
The notifications we were sending to attendees listed the event's _end_ time as the checkin time.

### DIFF
--- a/panels/models/attraction.py
+++ b/panels/models/attraction.py
@@ -663,16 +663,32 @@ class AttractionEvent(MagModel):
         return 'unknown start time'
 
     @property
-    def checkin_time(self):
+    def checkin_start_time(self):
         advance_checkin = self.attraction.advance_checkin
         if advance_checkin < 0:
-            return self.end_time
+            return self.start_time
         else:
             return self.start_time - timedelta(seconds=advance_checkin)
 
     @property
-    def checkin_time_label(self):
-        checkin = self.checkin_time_local
+    def checkin_end_time(self):
+        advance_checkin = self.attraction.advance_checkin
+        if advance_checkin < 0:
+            return self.end_time
+        else:
+            return self.start_time
+
+    @property
+    def checkin_start_time_label(self):
+        checkin = self.checkin_start_time_local
+        today = datetime.now(c.EVENT_TIMEZONE)
+        if checkin.date() == today.date():
+            return checkin.strftime('%-I:%M %p')
+        return checkin.strftime('%-I:%M %p %a')
+
+    @property
+    def checkin_end_time_label(self):
+        checkin = self.checkin_end_time_local
         today = datetime.now(c.EVENT_TIMEZONE)
         if checkin.date() == today.date():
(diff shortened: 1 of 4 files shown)


panels/models/attraction.py
```diff
@@ -663,24 +663,40 @@ def start_time_label(self):
...
+        advance_checkin = self.attraction.advance_checkin
+        if advance_checkin < 0:
+            return self.end_time
+        else:
+            return self.start_time
+
+    @property
+    def checkin_start_time_label(self):
+        checkin = self.checkin_start_time_local
+        today = datetime.now(c.EVENT_TIMEZONE)
```
EliAndrewC: I wouldn't hold up this PR based on this minor stylistic nitpick, but ``today`` intuitively sounds like a ``datetime.date`` object, whereas ``now`` sounds more like a ``datetime.datetime`` object.

EliAndrewC (approved):
+1

The new property names ``checkin_start_time`` and ``checkin_end_time`` make it much clearer what's going on - good stuff!

panels/models/attraction.py
```diff
@@ -663,24 +663,40 @@ def start_time_label(self):
...
 
     @property
-    def checkin_time_label(self):
-        checkin = self.checkin_time_local
+    def checkin_end_time(self):
+        advance_checkin = self.attraction.advance_checkin
+        if advance_checkin < 0:
+            return self.end_time
+        else:
+            return self.start_time
```
kitsuta, line 679: Isn't this a problem for the attractions that want to NOT let attendees check in right beforehand? I thought the point of advanced check in time was "you must check in AT LEAST XX minutes beforehand to get in"

EliAndrewC: Unless I've misunderstood (please correct me if I have) those attractions would configure things so that attendees must check in XX minutes beforehand, right?

RobRuana: The notifications will give the desired checkin, but if someone is late, and the event hasn't started yet, I wanted the event to still be able to check people in. Which I guess they could anyway, but this seemed like a safe option.

EliAndrewC: Oh wait, I think I see what you're saying - basically that under this logic, no matter what the XX minutes beforehand that's specified, all attractions would still allow people to check in 1 minute before the attraction.

Is that not okay?  I guess it's not desirable, but if we wanted to do otherwise then we'd need a much bigger change.  Because as of right now we only have a setting for each feature that says how many minutes beforehand checkin STARTS.  We would need a whole other database column that was like, "And here's how long before the event the checkin period lasts" which we don't have right now.

Unless I'm misunderstanding, which I certainly may be.

RobRuana: You have it right

RobRuana:
I actually do feel like your suggestion relieves a tiny amount of mental burden. So I made the change!

